PR #5, opened 2018-11-23
Add Minecraft Java Edition

moosad:
I know it's not freeware, but actually you just need a licensed account to login, the launcher is actually available to everyone. So why couldn't we install via scoop?

Also it was requested: https://github.com/lukesampson/scoop-extras/issues/1373


Calinou:
Thanks! :smiley:

PR #8, opened 2018-12-06
Add taisei v1.2

MCOfficer:
Please check the license, not 100% confident about it


Akaricchi:
It's the [MIT license](https://en.wikipedia.org/wiki/MIT_License), also known as the expat license. The assets are mostly CC-BY, some are CC-0.

Calinou:
Thanks! :smiley:

PR #9, opened 2018-12-10
Add standard scoop maintenance tools and tests

r15ch13:
This will allow me to add this bucket to the [Excavator](https://github.com/r15ch13/Excavator) Service


Calinou:
Thanks! :smiley:

PS: I added you as a collaborator to this repository so that it can stay maintained in case I become unavailable for any reason.

PR #10, opened 2018-12-15
Add WitchBlast v0.7.5

MCOfficer:


witchblast.json
```diff
@@ -0,0 +1,25 @@
...
+		"ad8faa4910abe1bd1cc11036b4c0d9956621224473a13a086230c5750cd3226f"
+	],
+    "extract_dir": "Witch Blast v0.7.5",
+    "bin": [
+		[
+			"run_working_dir.cmd",
+			"WitchBlast",
+			".",
+			"\"Witch Blast.exe\""
+		]
```
MCOfficer: this is ugly, but according to lukesampson/scoop#109 it's the only way to get it running

witchblast.json
```diff
@@ -0,0 +1,25 @@
...
+	],
+    "checkver": "github",
+	"##": "Note: No Autoupdate due to inconsistent releases, downloads etc. Also, a new release seems unlikely"
```
MCOfficer: please tell me if you want comments like this or not

MCOfficer:
Shortcuts look very promising - learned something new today. Too bad they don't fix this for CLI tools, but for games it'll do.

Also, thanks for the persists.

PR #13, opened 2019-01-28
Add pychess v0.99.4 

kiedtl:
Add the chess client [pychess](https://github.com/pychess/pychess) v0.99.4.


Calinou:
Thanks! :slightly_smiling_face:

PR #23, opened 2019-05-20
Add cataclysm-dda and cataclysm-dda-tiles.

matthewjberger:
This adds the open source roguelike Cataclysm: Dark Days Ahead curses and graphical versions.

I didn't include an autoupdate for these manifests because the version for releases are non-semver and scheduled for 6-month intervals (https://cataclysmdda.org/releases/)


Calinou:
Thanks! :smiley:

I amended the pull request to add a description and removed the `bin` property as it wasn't working when I tested it. I also tweaked the JSON formatting to follow other manifests.

matthewjberger:
The bin properties worked for me for both manifests. I ran them from cmder which is what I have been playing the ascii version in via the bin property

PR #25, opened 2019-06-08
Adds Summoner's War Exporter (current version: 0.0.25)

mmichaelis:
Summoner's War Exporter is a Proxy which will capture your game data for Summoners War: Sky Arena, which may then be used to optimize runes or to manage monsters at online platforms such as https://swarfarm.com/.

Summoners War: Sky Arena, is a mobile turn-based strategy MMO game for Android and iOS.

See Also:

* https://swarfarm.com/
* https://github.com/Xzandro/sw-exporter


Calinou:
Thanks!

PR #31, opened 2019-06-23
Add VisualBoyAdvance-M (and add notes on the VisualBoyAdvance manifest)

issaclin32:
VisualBoyAdvance-M is a community-maintained, improved fork of the original VisualBoyAdvance.
Homepage: https://github.com/visualboyadvance-m/visualboyadvance-m

Please let me know if there are any problems. Thanks!


Calinou:
If this is merged, we should probably deprecate `visualboyadvance` entirely (i.e. removing it after a few weeks have passed), as it's usually better to rely on maintained software.

issaclin32:
> If this is merged, we should probably deprecate `visualboyadvance` entirely (i.e. removing it after a few weeks have passed), as it's usually better to rely on maintained software.

Agree.
If there are no specific reason to keep the original `VisualBoyAdvance`, I think it would be better to only keep the `VisualBoyAdvance-M` package.

Calinou:
Thanks! :smiley:

PR #32, opened 2019-06-30
Add Minecraft-related tools: Amidst, Minutor, MCEdit2

issaclin32:
Homepage of **MCEdit2**: https://www.mcedit.net/
Homepage of **Amidst**: https://github.com/toolbox4minecraft/amidst
Homepage of **Minutor**: http://seancode.com/minutor/

Please let me know if there are any problems. Thanks.


Calinou:
Thanks! :smiley:

I made some changes so that Amidst runs correctly (it needs a Java installation). I also based it on the latest beta version instead of the stable release, as the stable release only supports Minecraft 1.11 (which is quite old by now).

PR #33, opened 2019-07-12
0ad: Add version 0.0.23b

issaclin32:
**0 A.D.** ([homepage](https://play0ad.com/)) is a free, open-source game of ancient warfare.

Notes:
* This package installs the files by **running the installer**. Is this acceptable in the game bucket?

I know that in the general criteria of scoop, it is preferred to extract the files manually (rather than running the installer). The installer seems to be created by [this NSIS script](https://github.com/0ad/0ad/blob/master/source/tools/dist/0ad.nsi), but I could not find a way to extract it using 7zip/uniextract/innounp.

Therefore, I followed [this example from the nonportable bucket](https://github.com/oltolm/scoop-nonportable/blob/master/notepadplusplus-np.json) to create the manifest. The files will be installed under the scoop directory, and the shortcuts are created by the installer itself.

* The file name of the installer is `0ad-0.0.23b-alpha-win32.exe`. I tend to think `-alpha` is not a part of the **version identifier** (it is just stating that the game is still in development), since `-alpha` exists in the filename of every releases.

### DIFF
--- a/bucket/0ad.json
+++ b/bucket/0ad.json
@@ -1,0 +1,25 @@
+{
+    "homepage": "https://play0ad.com/",
+    "description": "A free, open-source game of ancient warfare",
+    "version": "0.0.23b",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://github.com/0ad/0ad/blob/master/LICENSE.txt"
+    },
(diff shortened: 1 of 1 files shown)


Calinou:
Thanks!

PR #41, opened 2019-07-27
Add Blitz

### LINKED ISSUE #12: Add Blitz.gg
[Blitz.gg](https://blitz.gg/) is an app that helps find runes and builds in [League of Legends](https://leagueoflegends.com/)'s champion selection phase. Its installer is an archive containing a `.nupkg`, which I've heard can be used by Scoop.

issaclin32:
Close #12.

**Blitz** ([homepage](https://blitz.gg/)) is an app that helps find runes and builds in League of Legends's champion selection phase.

Notes:
* I couldn't find any page that contains the current version number of **Blitz**. Therefore the only way to add this package is through **nightly versions**. Is this acceptable? (or do you prefer not to add this package into the *games* bucket?)

* **Blitz** will be running in the background (showing in the tray) after the main window is closed. The command in the *uninstall* section avoids error when the user tries to uninstall the package while Blitz is still running in the background.

### DIFF
--- a/bucket/blitz.json
+++ b/bucket/blitz.json
@@ -1,0 +1,28 @@
+{
+    "homepage": "https://blitz.gg/",
+    "description": "App that helps find runes and builds in League of Legends's champion selection phase",
+    "version": "nightly",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://blitzesports.com/tos"
+    },
+    "url": "https://dl.blitz.gg/download/windows#/dl.7z",
+    "depends": "7zip",
+    "installer": {
+        "script": [
+            "7z x \"$dir\\Blitz-*-full.nupkg\" -o\"$dir\" lib\\net45",
+            "Move-Item -Path \"$dir\\lib\\net45\\*\" -Destination \"$dir\" -Force",
+            "Remove-Item -Path \"$dir\\Blitz-*-full.nupkg\""
+        ]
+    },
+    "uninstaller": {
+        "script": "taskkill /F /IM Blitz.exe /FI \"status eq running\""
+    },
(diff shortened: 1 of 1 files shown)


Calinou:
Thanks! :smiley:

> - I couldn't find any page that contains the current version number of Blitz. Therefore the only way to add this package is through nightly versions. Is this acceptable? (or do you prefer not to add this package into the games bucket?)

Nightly builds are fine if no other options are available (or as an additional package with a `-nightly` suffix, if a stable release is already available).

> - Blitz will be running in the background (showing in the tray) after the main window is closed. The command in the uninstall section avoids error when the user tries to uninstall the package while Blitz is still running in the background.

I don't think we can do much about it, so let's leave it as is.

PS: I amended your commit to tweak the description for consistency with other packages.

NatoBoram:
They have some sort of API, but it's not documented, quite a hassle to use, it changes sometimes, it doesn't really work and it displays unexpected information.

I found this : https://dl.blitz.gg/update/channel/stable/_/0.9.9

It shows information about the latest version, `1.1.4`.

If you explore a bit the URLs, you'll find out most endpoints just don't work. I wouldn't rely on their API, personally.

r15ch13:
@NatoBoram They use [Squirrel](https://github.com/Squirrel/Squirrel.Windows) for installing/updating and I [changed the manifest](https://github.com/Calinou/scoop-games/commit/e27498b77f69e17ed1ef85714716fefdd6dae83a) to use the same endpoints they use.

NatoBoram:
I asked about their API in their Discord, and I got this response : 

> The API is meant for "internal" usage only and it works the way we need it to.

PR #43, opened 2019-07-29
Add NetHack

sigvei:


sigvei:
Sorry, this manifest doesn't handle save files very well. I will open a new pull request when that is sorted out.

PR #46, opened 2019-08-05
Add League of Legends

Harakku:
There doesn't seem to be easy ways to extract BitRock installers but the pre- and
post-installation scripts make the experience not too bad.

The goal of the pre-installation script is to make sure that the installation is ran with
admin rights and that the installation cleanly exists otherwise.

The uninstaller sometimes threw errors when it was ran as _file+args_. Not sure
why that happens but when it's ran as _script_ with "start-process -wait" the occasional
errors seem to disappear.

### DIFF
--- a/bucket/leagueoflegends.json
+++ b/bucket/leagueoflegends.json
@@ -1,0 +1,50 @@
+{
+    "##": [
+        "BitRock installer. Requires admin privileges. Some of the unwanted behavior fixed in the post_install",
+        "Probably wanted behavior: adds a firewall rule.",
+        "Unwanted behavior: creates desktop and start menu shortcuts, adds a registry key for 'Programs and Features' uninstallation."
+    ],
+    "version": "nightly",
+    "homepage": "http://leagueoflegends.com/",
+    "description": "MOBA game published by Riot Games.",
+    "license": {
+        "identifier": "Freeware",
+        "url": "https://euw.leagueoflegends.com/en/legal/termsofuse"
+    },
+    "url": "https://riotgamespatcher-a.akamaihd.net/releases/live/installer/deploy/League%20of%20Legends%20installer%20EUW.exe#/installer_euw.exe",
+    "persist": [
+        "Config"
+    ],
+    "shortcuts": [
+        [
+            "LeagueClient.exe",
+            "League of Legends"
+        ]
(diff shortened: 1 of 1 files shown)


Calinou:
Installating and uninstalling seem to work here. I haven't tested logging in and playing though (I don't play League of Legends :stuck_out_tongue:).

I noticed that the download hash can change without notice:

```text
App:         leagueoflegends
URL:         https://riotgamespatcher-a.akamaihd.net/releases/live/installer/deploy/League%20of%20Legends%20installer%20EUW.exe#/installer_euw.exe
First bytes: 4D 5A 90 00 03 00 00 00
Expected:    e309fd9e276526385ee53a8caf4f8121f022148117b2e2c1f48c8e31c93a15af
Actual:      3bd13850bfd10ee8fa4a83de3e4aac386a88053c3edd4bc4770d77d3dd1d3803
```

If this is merged, we'll need to mark the package version as `nightly` and remove the `hash` property.

@r15ch13 Do you know if there's a way to run specific commands with elevated privileges in PowerShell? Could [`sudo`](https://github.com/ScoopInstaller/Main/blob/master/bucket/sudo.json) be used for this?

Harakku:
I changed it to nightly and removed the "Programs and Features" hack. The latter can probably be handled in a general way by targeting registry keys with scoop paths in the values.

Some kind of UAC tag in the manifest would make sense. I dunno if there exists some talk about this already.

r15ch13:
@Calinou Yes, you can use [Invoke-ExternalCommand](https://github.com/lukesampson/scoop/blob/79f8538b57b9021db71a279879b9032fefd1ae52/lib/core.ps1#L375) with the `-RunAs` flag.

Calinou:
@r15ch13 I tried replacing the `installer` section with this section to use `Invoke-ExternalCommand`, but I get an :

```json
    "installer": {
        "script": "Invoke-ExternalCommand -FilePath \"$dir\\installer_euw.exe\" -ArgumentList \"--mode unattended --installdir `\"$dir`\" --license_agreement 1\" -RunAs"
    },
```

```text
ERROR Exception calling "Start" with "0" argument(s): "The requested operation requires elevation"
```

It works when using an elevated command prompt, so the command itself is OK.

PR #61, opened 2019-10-13
Add gog-galaxy-plugin-downloader

JanPokorny:


Calinou:
Like #62, I think this would be better suited to [scoop-extras](https://github.com/lukesampson/scoop-extras) as it already contains Steam-related tools.

The idea is to keep this bucket "topical", as it's about *games* and their related tools. Sorry if this is unclear, I should probably write a CONTRIBUTING.md file someday.

JanPokorny:
@Calinou So the idea is that this bucket is primarily for games, and secondarily for tools related to a specific game, like map editors? Maybe include a line like "Game tools should relate to a specific game. General gaming tools and utilities should go to scoop/extras." in the README.

Calinou:
@JanPokorny Exactly :slightly_smiling_face:

JanPokorny:
Moved to https://github.com/lukesampson/scoop-extras/pull/2975.

PR #62, opened 2019-10-13
Add XJoy

JanPokorny:


Calinou:
I think this should be submitted to [scoop-extras](https://github.com/lukesampson/scoop-extras) instead, as other drivers like DS4Windows are already there :slightly_smiling_face:

JanPokorny:
OK, moved to https://github.com/lukesampson/scoop-extras/pull/2974

PR #66, opened 2019-11-11
Tweaks to minecraft.json

TheRandomLabs:
Feedback is welcome. Here are the changes I made:
* `version` has been changed to `nightly` so that the hash doesn't have to be manually updated.
* `license.identifier` has been changed to `Proprietary` as the game is not free (although I suppose the *launcher* is free).
* Using the `current` junction no longer causes any issues, at least for me.
* Added `game` and `runtime` to persist. This is because I use `scoop update minecraft --force --no-cache` to update Minecraft, and redownloading `game` and `runtime` every time the launcher is updated is time-consuming.
:)

### DIFF
--- a/bucket/minecraft.json
+++ b/bucket/minecraft.json
@@ -1,20 +1,22 @@
 {
+    "version": "nightly",
+    "description": "Official launcher for Minecraft, a sandbox voxel game.",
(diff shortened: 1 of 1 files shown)


Calinou:
Thanks!

PR #69, opened 2019-11-15
Add dolphin stable and dolphin dev

BlackPowerade:
Was surprised to not find dolphin here.
These are mostly poached from the citra manifests.


BlackPowerade:
Is there anything else that should be done to these? I assume it is ok to leave the architecture out since neither stable nor dev support anything but 64 bit.

Calinou:
Thanks! It looks good to me, sorry for the delay :slightly_smiling_face:

PR #70, opened 2019-11-24
Add melonDS (initial)

BlackPowerade:


BlackPowerade:
Does anyone know how to add color to the notes section? I would like to add some color to make the ATTENTION standout among the other output text.

Calinou:
Thanks!

> Does anyone know how to add color to the notes section?

Would something like this work? https://devblogs.microsoft.com/scripting/powertip-write-powershell-output-in-color-without-using-write-host/

PR #71, opened 2019-11-26
Add PCSX2 and PCSX2-dev

BlackPowerade:
For some reason $matchRandom wouldn't work with an & immediately after it, and I couldn't figure out how to escape it.
Fortunately the git revision appended to each build is always 10 characters long.

### DIFF
--- a/bucket/pcsx2-dev.json
+++ b/bucket/pcsx2-dev.json
@@ -1,0 +1,45 @@
+{
+    "homepage": "https://pcsx2.net/",
+    "description": "A feature rich FOSS PlayStation 2 emulator (development version)",
+    "license": "LGPL-3.0-or-later",
+    "version": "1.5.0-3315",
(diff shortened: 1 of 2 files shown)


matoro:
Glad these emulators are finally getting into the main bucket.  I have been running a `pcsx2-nighly` package in my bucket for a while now: https://github.com/matoro/bucket/blob/master/pcsx2-nightly.json , not sure if you might want to incorporate some of my choices
- `vcredist2015` in `depends`
- `post_install` script that disables the BIOS prompt screen, this might make sense on nightly flavor only otherwise you have to click through that screen each time you update
- should `shortcuts` for the nightly flavor have a different name, otherwise I think it will conflict if you try to install both at the same time

BlackPowerade:
Thanks @matoro , I added vcredist2015 and vcredist to `suggests` since scoop can't seem to handle mutually exclusive depends yet. Also changed up the persistence scheme, since some things shouldn't be persisted. Persisting portable.ini should prevent the wizard from starting after an update.
I also changed the shortcut to PCSX2-dev.

Calinou:
Thanks!

PR #73, opened 2019-12-04
Warzone2100 installation fixed

jfastnacht:
There was a problem with the Warzone2100 installation, since the release URL seem to have changed.


Calinou:
Thanks! I tested it locally and it works. :+1:

PR #74, opened 2019-12-07
Citra(-canary): add persistence

BlackPowerade:
activates Citra's portable mode and migrates the user files from `%APPDATA%`

### DIFF
--- a/bucket/citra-canary.json
+++ b/bucket/citra-canary.json
@@ -2,10 +2,20 @@
     "homepage": "https://citra-emu.org/",
     "description": "Emulator for Nintendo 3DS, capable of playing many of your favorite games",
     "license": "MIT",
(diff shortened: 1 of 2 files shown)


Calinou:
Why not use files from `APPDATA`? It seems like it's less trouble to use the default storage mechanism rather than relying on portable mode and a persistent directory.

BlackPowerade:
Mostly to give it more functionality with scoop, and prevent directory pollution. Kind of a pain to go searching for config files when some scoop packages keep them in `persist` and some keep them in `APPDATA`
It also allows for seperate configs between canary and stable, since both of them read from `APPDATA` primarily.

Calinou:
Alright, I'll merge this then :slightly_smiling_face:

Thanks!

PR #76, opened 2019-12-12
Add persist for BGB, Kega-Fusion, Snes9x packages

### LINKED ISSUE #75: No persist folders
BGB, Kega-Fusion and Snes9x don't have persist folders.
Please add them.

issaclin32:
close #75

This adds **persist** to `BGB`, `Kega-Fusion` and `Snes9x` packages.

* `BGB` already have its config file (bgb.ini) in the archive, therefore we don't need to create a new file in the **pre_install** section.

* I have tested the packages to make sure that `Kega-Fusion` and `Snes9x` works properly (will not crash) with an empty config file.

### DIFF
--- a/bucket/bgb.json
+++ b/bucket/bgb.json
@@ -27,6 +27,7 @@
             ]
         }
     },
+    "persist": "bgb.ini",
     "checkver": {
         "url": "http://bgb.bircd.org/",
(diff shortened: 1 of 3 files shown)


Calinou:
Thanks!

PR #79, opened 2019-12-25
dolphin-dev: Fix autoupdate

BlackPowerade:
For some reason the url scheme changed and some random characters were added.


Calinou:
Thanks!

PR #80, opened 2019-12-26
mesen: Add version 0.9.8

wenmin92:
[Mesen](https://www.mesen.ca/): a high-accuracy NES and Famicom emulator and NSF player for Windows and Linux.

Github: https://github.com/SourMesen/Mesen

**Note:**
It's data can be saved into the user's document folder or application folder. If user download the zip, extract and execute the mesen.exe, will let user select which folder to save data. Here I force save data in the appdata by create a settings.xml file.


Calinou:
Thanks!

PR #84, opened 2019-12-28
sw-exporter: Update to version 0.0.31 (manually)

issaclin32:
**Summoner's War Exporter** has changed their file name pattern.  Therefore the package cannot be automatically updated by the *Excavator*. This fixes the issue.


Calinou:
Thanks!

PR #85, opened 2019-12-28
nethack: Update to version 3.6.4 (manually)

issaclin32:
**nethack** has changed their file name pattern. Therefore the package cannot be automatically updated by the Excavator. This fixes the issue.

I modified the manifest to download file from the Github repo instead of `nethack.org` because I cannot find a proper way to parse the pages to get the correct URL from `nethack.org`. 
(See https://www.nethack.org/common/index.html for details)


Calinou:
Thanks!

PR #86, opened 2019-12-28
gzdoom, lzdoom: Update manually

issaclin32:
**gzdoom** and **lzdoom** has changed their file name pattern. Therefore the packages cannot be automatically updated by the Excavator. This fixes the issue.


Calinou:
Thanks!

Issue #91, opened 2020-03-01
Add modorganizer


iovw:
A mod manager from https://github.com/ModOrganizer2/modorganizer

PR #92, opened 2020-03-01
add mo2

iovw:
#91

### DIFF
--- a/bucket/mo2.json
+++ b/bucket/mo2.json
@@ -1,11 +1,11 @@
-﻿{
-    "homepage":  "https://github.com/ModOrganizer2/modorganizer",
+{
+    "homepage": "https://github.com/ModOrganizer2/modorganizer",
     "description": "Mod manager for various PC games",
-    "version":  "2.2.2.1",
-    "license":  "GPL-3.0",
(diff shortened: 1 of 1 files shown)


iovw:
Is this ok now? @Calinou

bucket/mo2.json
```diff
@@ -0,0 +1,21 @@
+{
+    "homepage": "https://github.com/ModOrganizer2/modorganizer",
+    "description": "Mod manager for various PC games",
+    "version": "2.2.2.1",
+    "license": "GPL-3.0",
+    "url": "https://github.com/ModOrganizer2/modorganizer/releases/download/v2.2.2.1/Mod.Organizer-2.2.2.1.7z",
+    "hash": "sha256:65B254431AC5F3BA23251694D8F7B8910243A412EE9D8AEDE1C3259705565FAF",
```
Calinou: The default hash format is SHA-256, no need to specify it. Also, we use lowercase as per convention:

```suggestion
    "hash": "65b254431ac5f3ba23251694d8f7b8910243a412ee9d8aede1c3259705565faf",
```

iovw: I mean I think explicit is better than implicit.

Calinou:
Thanks!

Issue #94, opened 2020-03-04
Adding OpenMW


apommel:
OpenMW is a popular Open Source reimplementation of The Elder Scrolls III: Morrowind engine, and is still under heavy development (you can check out their [github](https://github.com/OpenMW/openmw)).

I tried to write [a manifest](https://github.com/apommel/scoop-bucket-build/blob/master/openmw.json) to make your work easier, but I don't know how to deal with persist data (saves and configuration files), which seems to be currently hardcoded to go to `C:\Users\username\Documents\My Games\OpenMW` ([ref1](https://forum.openmw.org/viewtopic.php?t=5097), [ref2](https://forum.openmw.org/viewtopic.php?f=2&t=3402#p38007)), so you can take my manifest as a reference but if you can improve that it would be great. Also I currently added the same shortcuts as their installer, but it may be a lot, I'm not sure. Also, the entire installer content is currently extracted but some parts are not useful (installer data in `$PLUGINSDIR` and uninstaller executable), I don't know if that would be an issue.

Also it would be nice to have an `openmw-nightly` app as the nightlies are actually used a lot (they are available [here](https://forum.openmw.org/viewtopic.php?t=1808) along with their hash), but I don't know how to deal with the versioning.

Thanks in advance!

Calinou:
> I tried to write a manifest to make your work easier, but I don't know how to deal with persist data (saves and configuration files), which seems to be currently hardcoded to go to C:\Users\username\Documents\My Games\OpenMW (ref1, ref2), so you can take my manifest as a reference but if you can improve that it would be great.

If the application saves its data outside the program directory, then you don't need to do anything specific to handle it :slightly_smiling_face:

apommel:
I believe the pull request resolved this issue!

PR #95, opened 2020-03-04
Add OpenMW stable and nightly

apommel:
This PR adds the manifest for `openmw` and `openmw-nightly` packages ( #94 ). Tell me if anything seems wrong.


Calinou:
Thanks for your contribution :slightly_smiling_face: 

I noticed nothing happens when starting OpenMW for the first time (no error dialog appears). Instead, I need to use the launcher or wizard to see some kind of window appear (telling me to install data files, as expected). Should we remove the "direct" OpenMW option from the start menu? It could confuse users as they're expected to go through the launcher first.

We could likely also remove the OpenMW Wizard start menu entry as people will only need to use it once, and it's already referenced in the Launcher.

bucket/openmw.json
```diff
@@ -0,0 +1,42 @@
+{
```
Calinou: This file should use 4 spaces for indentation (like all other manifests).

apommel: I think this is because I used `scoop create` to generate this manifest. I correcte dit now, thanks.

apommel:
I am not sure about removing the "direct" entry as after the config files have been generated with the launcher, it is the main way of launching the game. Do you think that adding a `notes` specifying that the user should launch the launcher first would be enough?

For the remaining parts, I agree with you. Do you think that would also be helpful to rename the shortcuts of the nightly (for example `OpenMW` -> `OpenMW (nightly)` ) in case both versions are installed?

apommel:
I realized that the `checkver` had some issues, even with the github one. It was fixed with the latest commit.

Calinou:
> I am not sure about removing the "direct" entry as after the config files have been generated with the launcher, it is the main way of launching the game. Do you think that adding a notes specifying that the user should launch the launcher first would be enough?

Yes, a `notes` property should suffice. Something like:

`Please run the OpenMW Wizard in the start menu to configure the game data path. Otherwise, OpenMW won't start correctly.`

apommel:
> We should also find a way to rename the shim executables with a `-nightly` suffix so they don't conflict with the stable version. This way, people can install both the stable and nightly versions at the same time.

I was wondering about that too, I did it the [python27 way](https://github.com/ScoopInstaller/Versions/blob/master/bucket/python27.json). Everything should have been fixed in the latest commit.

Calinou:
Thanks!

PR #97, opened 2020-03-05
lf2: Add version 2.0a

issaclin32:
**Little Fighter 2** (often abbreviated as **LF2**) is a free fighting game that is well-known in the Chinese-speaking region.  ([homepage](http://www.lf2.net/)) ([Wikipedia](https://en.wikipedia.org/wiki/Little_Fighter_2))

Notes:
* *checkver/autoupdate* is not needed because **LF2** has not been updated since 2009.

* The installer from the official source cannot be properly extracted. Also the silent switch (`/S`) causes error for some reason. Therefore I use a **third-party source** ([azo freeware](https://www.azofreeware.com/2011/12/little-fighter-20a.html)) instead. Is this acceptable?

### DIFF
--- a/bucket/lf2.json
+++ b/bucket/lf2.json
@@ -1,0 +1,17 @@
+{
+    "version": "2.0a",
+    "description": "Little Fighter 2. A Fighting game.",
+    "homepage": "http://www.lf2.net/",
+    "license": "Freeware",
+    "url": "http://201708.mediafire.com/file/j2ddn2qqc27u5x1/LittleFighter_2.0a_azo.exe/file#/dl.7z",
+    "hash": "254c79d8fb0af8df72eef042aa2f0c3d9409a0da391f6ab40964450fb0ade034",
+    "extract_dir": "LittleFighter",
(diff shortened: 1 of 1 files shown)


Calinou:
I'm getting an error on startup on Windows 10:

image

I've found [a post about that error message online](https://answers.microsoft.com/en-us/windows/forum/windows_10-other_settings/windows-10-error-the-application-has-failed-to/5c6a015e-4162-491d-b7d9-f3b13612e304) but I'm not sure how well this applies to an application released in 2009.

issaclin32:
> I'm getting an error on startup on Windows 10:

I found that this problem can be solved by installing **VCRedist 2005**. (`scoop install vcredist2005`).

Also there will be an error if we start the application(`lf2.exe`) without setting *working directory* to the same path as `lf2.exe`. Therefore I added a PS-script so that **shim-exes** can work properly.

Thank you for testing it out!

Calinou:
I still can't get it to work after installing `vcredist2005`, but I'll merge it anyway as I suppose it works on your PC.

PR #98, opened 2020-03-12
Zenith: Add version 0.9.10.r1

abejenaru:


Calinou:
I'm getting an error on startup on Windows 10. It might be due to me using a virtual machine without Direct3D support though. (I have OpenGL 3.3 support thanks to [mesa-dist-win](https://github.com/pal1000/mesa-dist-win).)

image

I'll try to run it on bare metal when I find some time and report if it works.

abejenaru:
My bad, I only tested it on Windows 7 and didn’t encounter any issue. On the game’s page, I see it requires OpenAL!

PR #99, opened 2020-03-12
Augmentinel: Add version 1.3

abejenaru:


Calinou:
Thanks for opening a pull request :slightly_smiling_face: 

Unfortunately, the program fails to start if DirectX libraries aren't installed on the system:

image

Since we can't install them automatically using Scoop, I suppose this should be mentioned in the `notes` field like this:

```text
"notes": "If you get an error message about X3DAudio1_7.dll being missing, try installing the DirectX end-user package: https://www.microsoft.com/en-us/download/confirmation.aspx?id=8109"
```

Calinou:
Thanks!

PR #102, opened 2020-04-06
Fix version number for openttd

fbartels:
When the installation files of openttd were updated the version number itself was forgotten.


Calinou:
Thanks!

PR #103, opened 2020-04-12
Add mednafen(-dev) and mednaffe manifests

### LINKED ISSUE #96: Add mednafen and possibly the mednaffe gui
mednafen is a command line multi-system emulator utilizing OpenGL and SDL.
https://mednafen.github.io

Typically one does not use it on its own thus there is also third party graphical user interface.
https://github.com/AmatCoder/mednaffe

BlackPowerade:
closes #96 
The mednaffe manifest is very convoluted. Since mednafen comes in 2 channels, dev and normal, I can't just put mednafen as a dependency. Personally I would like to end the mednaffe install if mednafen or mednafen-dev are not present or maybe prompt the user to install one or the other during the mednaffe install.
Any help with this would be greatly appreciated. Currently works but just warns and is real messy.


BlackPowerade:
After almost a year of forgetting about this PR and bothering me everytime I submit another, it finally comes time to put it to rest.     
These changes should stop the mednaffe install if no existing mednafen is found, and successfully persist keep the mednafen config in sync with the one inside mednaffe's folder (since it does not look anywhere besides where it resides for the config file.

bucket/mednaffe.json
```diff
@@ -0,0 +1,53 @@
...
+            "elseif (Test-Path \"$scoopdir\\apps\\mednafen\\current\\mednafen.exe\") {",
+            "   $medinst = 2",
+            "}",
+            "if ((Test-Path \"$scoopdir\\apps\\mednafen\\current\\mednafen.exe\") -and (Test-Path \"$scoopdir\\apps\\mednafen-dev\\current\\mednafen.exe\")) {",
+            "   Write-Host -ForegroundColor Yellow \"Mednafen and Mednafen-dev were found on this system.\"",
+            "   Write-Host -ForegroundColor Yellow \"Choose one to use mednaffe with, 1 (stable) or 2 (dev)\"",
+            "   $medinst = Read-Host -Prompt \"[1 / 2]\"",
+            "}",
+            "switch ( $medinst ) {",
+            "   0 { Write-Warning \"Mednaffe must be installed AFTER mednafen or mednafen-dev. Install either of them, then install mednaffe!\"",
```
Calinou, line 30: This message should appear as an error rather than a warning, since the execution is aborted just after printing it.

Calinou (commented):
This works locally, but I noticed that Mednaffe's install script will print an error message about `sudo` not being found if you didn't install it. (This only occurs if you already have `mednafen` or `mednafen-dev` installed.)

The installation is still successful, but it's strange to see an error message in this case.

bucket/mednaffe.json
BlackPowerade: Ok, I've changed it to an error and moved it to `pre-install`. It will also uninstall mednaffe to prevent scoop from saying the install failed in `scoop list`

BlackPowerade:
@Calinou Merge?

Calinou (approved):
Thanks!

PR #108, opened 2020-04-29
ps3-system-software: Add version 4.86

matoro:
Also added as a suggestion to rpcs3-dev.
This does not automatically set RPCS3 up to use the firmware.
You still have to install it through the interface.  This
just downloads it for you and should notify you via a package
upgrade when a new version is available.

### DIFF
--- a/bucket/ps3-system-software.json
+++ b/bucket/ps3-system-software.json
@@ -1,18 +1,17 @@
 {
     "version": "4.86",
-    "url": "http://dus01.ps3.update.playstation.net/update/ps3/image/us/2020_0331_cf9cb4ba53a83ad557501417837c8de9/PS3UPDAT.PUP",
-    "hash": "14ad0a4716d2bdf096dab9ecf77b9b838cc64435db7925837a38979fa05f4012",
     "description": "The PlayStation 3 system software is the updatable firmware and operating system of the PlayStation 3.",
     "license": {
         "identifier": "Proprietary",
         "url": "https://doc.dl.playstation.net/doc/ps3-eula/ps3_eula_en.html"
     },
+    "url": "http://dus01.ps3.update.playstation.net/update/ps3/image/us/2020_0331_cf9cb4ba53a83ad557501417837c8de9/PS3UPDAT.PUP",
+    "hash": "14ad0a4716d2bdf096dab9ecf77b9b838cc64435db7925837a38979fa05f4012",
     "checkver": {
         "url": "https://www.playstation.com/en-us/support/system-updates/ps3/",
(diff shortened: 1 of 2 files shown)


Calinou:
Thanks!

Calinou:
There's an error on AppVeyor:

```text
      [-] ps3-system-software.json has 1 Error!
      [*] Error: String '$matchDlurl' does not validate against format 'uri'.
        [^] Line: C:\projects\scoop-games\bucket\ps3-system-software.json:16:28
        [^] Path: http://scoop.sh/draft/schema#/definitions/autoupdate/properties/url/Format
      [-] ps3-system-software.json 71ms
        Expected 0, but got 1.
        79:                         $validator.Errors.Count | should -be 0
        at <ScriptBlock>, C:\projects\scoop\test\Scoop-Manifest.Tests.ps1: line 79
```

matoro:
> There's an error on AppVeyor, not sure if it can be worked around:
> 
> ```
>       [-] ps3-system-software.json has 1 Error!
>       [*] Error: String '$matchDlurl' does not validate against format 'uri'.
>         [^] Line: C:\projects\scoop-games\bucket\ps3-system-software.json:16:28
>         [^] Path: http://scoop.sh/draft/schema#/definitions/autoupdate/properties/url/Format
>       [-] ps3-system-software.json 71ms
>         Expected 0, but got 1.
>         79:                         $validator.Errors.Count | should -be 0
>         at <ScriptBlock>, C:\projects\scoop\test\Scoop-Manifest.Tests.ps1: line 79
> ```

I saw that, but as far as I can tell that's actually an issue with the schema.  The problem is, I have to capture the whole target URL out of the page text.  The schema is validating that the autoupdate URL looks like a URL, e.g. starting with http, but since I'm just dumping the whole regex match in there rather than using it as part of a larger URL, it gets confused.  Am I reading this right?  Do you have the same interpretation?

Calinou:
@matoro That's what I thought as well. Now, I wonder if we can make the CI script ignore this particular file so I don't have to deal with CI failing on every commit :slightly_smiling_face:

matoro:
I don't think it will, the console log shows that all packages that weren't changed were skipped.  However, I guess this should be filed as an upstream issue since technically the schema is wrong?  I think modifying this file should do it:  https://github.com/lukesampson/scoop/blob/master/schema.json

matoro:
@Calinou pushed a fix contributed by @Ash258 that approaches it much better and doesn't require any schema changes.  Next step is seeing if I can put together a patch for rpcs3 to install this with a command-line flag!

Calinou:
Thanks!

matoro:
@Calinou Somehow this seems to have caused the `rpcs3-dev` package to stop getting auto-bumped by CI.  It's stuck on an old version and with the wrong hash, too.

Calinou:
cc @r15ch13, who maintains the automated update bot.

PR #112, opened 2020-05-06
vkquake: Add version 1.04.1

bryanburke:
[vkQuake](https://github.com/Novum/vkQuake) is a Quake 1 port based on QuakeSpasm that uses the Vulkan graphics API for rendering.

I copied the QuakeSpasm manifest and made some basic additions for `checkver` and `autoupdate`. I also added a note from the vkQuake repo about needing the MSVC 2019 redists installed.

### DIFF
--- a/bucket/vkquake.json
+++ b/bucket/vkquake.json
@@ -1,0 +1,80 @@
+{
+    "homepage": "https://github.com/Novum/vkQuake",
+    "description": "Quake 1 port using Vulkan instead of OpenGL for rendering, based on QuakeSpasm",
+    "version": "1.04.1",
+    "license": "GPL-2.0-or-later",
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/vkquake.json
```diff
@@ -0,0 +1,80 @@
...
+                "extract_dir": "vkquake-$version_win64"
+            },
+            "32bit": {
+                "url": "https://github.com/Novum/vkQuake/releases/download/$version/vkquake-$version_win32.zip",
+                "extract_dir": "vkquake-$version_win32"
+            }
+        }
+    },
+    "notes": [
+        "vkQuake requires the Visual C++ Redistributable for Visual Studio 2019:",
```
Calinou, line 60: You can add `"depends": "extras/vcredist2019"` to the manifest so you don't have to tell the user manually in a note.

bryanburke: Thanks for the suggestion. I didn't realize any Scoop buckets had the redists available. Addressed in latest commit.

Calinou:
Thanks!

PR #122, opened 2020-05-18
minecraft: Add autoupdate

TheRandomLabs:
I've tested this, and it [seems to work](https://github.com/TheRandomLabs/Scoop-Bucket/commit/70ab7a3a492fdb00bea86a97b130657de1ed72a2). Note that this is for the launcher, and not the actual game itself. :)


TheRandomLabs:
That might have fixed it. :)

PR #126, opened 2020-06-10
openra: Update to version 20200503

Windafort:


Calinou:
For future reference, this PR was superseded by #127.

PR #167, opened 2020-11-23
Add Super Smash Flash 2

schance995:


schance995:
I don't know if persistent data is possible. It looks like the data and config are stored in the same files.

Calinou:
@schance995 You can specify individual files to be persisted (rather than folders), it should work.

schance995:
Thanks, I will check it out when I have time.

Calinou:
The URL is no longer valid. This should be updated to 1.2.5 before it can be merged :slightly_smiling_face:

schance995:
Thanks for the notice.
Also, I found that ssf2 updates automatically and silently, so I don't think there is a need for persistent data.
I would put the version as nightly, but the download URL changes with every version. If an nightly app's manifest is updated, will scoop update that app?

Calinou:
> I would put the version as nightly, but the download URL changes with every version. If an nightly app's manifest is updated, will scoop update that app?

No, Scoop never updates `nightly` applications once they've been installed (unless you explicitly request it).

I guess we have to request the developer to add some kind of permalink to the latest stable version.

schance995:
I double checked the persist data, they're all in zlib compressed binary form so I don't think I could find out what to persist in a reasonble time.

PR #170, opened 2020-12-03
sauerbraten: Update to version 2020-edition

willcassella:
Cube 2: Sauerbraten was recently updated for the first time since 2013, this PR updates the manifest for that


Calinou (approved):
Thanks! I tested it locally, it works.

(I amended the pull request to use a lowercase hash for consistency with other manifests.)

PR #172, opened 2020-12-06
minecraft: Fix launch error

### LINKED ISSUE #146: Minecraft Launcher installation broken
This issue concerns the "minecraft@2.1.16101" app.

I have noticed some time ago that the minecraft launcher installation made by scoop is somehow broken. I've tested it on two different machines and across some versions of the app, but whenever I install minecraft via `scoop install minecraft`. the resulting launcher executable cannot be launched; instead I get the error "Unable to update the Minecraft Native Launcher". Googling this error message yields several results, which all somehow suggest re-installing your launcher installation (Example: https://www.auslogics.com/en/articles/fix-unable-to-update-minecraft-native-launcher/). Uninstalling and re-installing minecraft via scoop does not resolve the issue, however. For now, the only solution that I found was to `scoop uninstall minecraft` and then download the installer directly from the minecraft homepage. 

I'd be happy if I could keep managing all my apps, including minecraft, with scoop.

I tried this using Windows 10.

TheRandomLabs:
Resolves #146. Thanks for waiting, exams got in the way. 😄

### DIFF
--- a/bucket/minecraft.json
+++ b/bucket/minecraft.json
@@ -1,4 +1,4 @@
- {
+{
     "version": "2.2.327",
     "description": "Official launcher for Minecraft, a sandbox voxel game.",
     "homepage": "https://www.minecraft.net/",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #176, opened 2020-12-11
amulet-map-editor: Add version 0.7.2.8

earnestma:
- From lukesampson/scoop-extras#5282
- Closes lukesampson/scoop-extras#5281

### DIFF
--- a/bucket/amulet-map-editor.json
+++ b/bucket/amulet-map-editor.json
@@ -1,0 +1,30 @@
+{
+    "version": "0.7.2.8",
+    "description": "Minecraft world editor and converter that supports all versions since Java 1.12 and Bedrock 1.7",
+    "homepage": "https://www.amulet-editor.com",
+    "license": "Unknown",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Amulet-Team/Amulet-Map-Editor/releases/download/0.7.2.8/Amulet-v0.7.2.8-windows-x64.zip",
+            "hash": "b4dcaf42181019b22835b8fd1a50731b0d3fa2475b37f4ce0dc9af1a0fcaf96d",
+            "extract_dir": "Amulet"
+        }
+    },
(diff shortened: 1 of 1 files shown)


Calinou:
Should the manifest be named `amulet` or `amulet-editor` instead of `amulet-map-editor`?

bucket/amulet-map-editor.json
```diff
@@ -0,0 +1,31 @@
...
+    "homepage": "https://www.amulet-editor.com",
+    "license": "Unknown",
+    "architecture": {
+        "64bit": {
+            "url": "https://github.com/Amulet-Team/Amulet-Map-Editor/releases/download/0.7.2.8/Amulet-v0.7.2.8-windows-x64.zip",
+            "hash": "b4dcaf42181019b22835b8fd1a50731b0d3fa2475b37f4ce0dc9af1a0fcaf96d",
+            "extract_dir": "Amulet"
+        }
+    },
+    "bin": "amulet_app.exe",
```
Calinou, line 13: The app dumps its configuration and logs to the CWD when started from a command prompt, which is bad for usability. It's better to not add it to the `PATH` in this case.

```suggestion
    },
```

Calinou (changes requested):

bucket/amulet-map-editor.json
earnestma: > Also, should the manifest be named `amulet` or `amulet-editor` instead of `amulet-map-editor`?

I'm not sure. Probably `amulet-editor` or `amulet-map-editor` <https://www.amulet-editor.com/>

Instead of launching from command prompt, would a shortcut be better?

earnestma: image

I'm having trouble applying your suggestion, so I'll do it manually in a moment

Calinou (approved):
Thanks!

PR #184, opened 2021-01-09
Rework amidst.json

jcwillox:
Switches to using the `.exe` version of amidst, this provides a simpler manifest/install. It also adds a start menu icon, which the batch file used previously did not.

I also moved the version field to the top, based on the scoop formatting standards.

### DIFF
--- a/bucket/amidst.json
+++ b/bucket/amidst.json
@@ -1,23 +1,19 @@
 {
-    "homepage": "https://github.com/toolbox4minecraft/amidst",
-    "description": "Minecraft interface and data/structure tracking tool",
     "version": "4.6",
+    "description": "Minecraft interface and data/structure tracking tool",
+    "homepage": "https://github.com/toolbox4minecraft/amidst",
     "license": "GPL-3.0-only",
-    "url": "https://github.com/toolbox4minecraft/amidst/releases/download/v4.6/amidst-v4-6.jar#/amidst.jar",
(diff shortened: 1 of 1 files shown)


bucket/amidst.json
```diff
@@ -1,23 +1,19 @@
...
         [
-            "amidst.bat",
+            "amidst.exe",
             "Amidst"
         ]
     ],
-    "pre_install": "Set-Content -Path \"$dir\\amidst.bat\" -Value \"pushd $dir && javaw -jar amidst.jar && popd\"",
-    "suggest": {
-        "Java Runtime Environment": "java/adoptopenjdk-hotspot-jre"
-    },
```
Calinou, line 18: This part is still needed since the executable doesn't bundle a JRE. (If it did, it would be 60+ MB.)

jcwillox: Good catch, didn't realise that 👍

Calinou:
`amidst.exe` can't detect my OpenJDK installation from `java/adopt8-hotspot-jre`, I don't know why.

jcwillox:
Interesting, if that's the case I completely agree. I know it worked for me, maybe because I have Java installed both through `scoop` and the normal Java installer. I'll see if I can get this working on one of my other devices where I've only installed Java through `scoop`.

jcwillox:
I gave it a try and you're correct the program will not detect Java installed via `scoop`, so I'm going to close this PR for now. I'll probably try and update the current jar version to include an icon and support command line args instead, or open an issue on the amidst repo.

PR #188, opened 2021-02-07
openmw-nightly: Fix manifest for new build system

apommel:
A fix once again for the new build system for openmw-nightly. The manifest was update to the latest version in the process.


Calinou (approved):
Thanks!

PR #190, opened 2021-02-08
openmw-nightly: Fix for Windows PowerShell

apommel:
I realized that the manifest I made yesterday was only compatible with PowerShell Core, so I corrected it for Windows PowerShell and made it a little bit cleaner in the process. Sorry for the trouble.


Calinou (approved):
Thanks!

PR #201, opened 2021-02-15
openmw-nightly: Fix for new archive filename

apommel:
A small fix as the way the files were named changed slightly with a commit yesterday.


Calinou (approved):
Thanks!

PR #206, opened 2021-02-21
mgba-dev: return to normal versioning

BlackPowerade:
It looks like at some point in time (or maybe I just missed it when I made it) mgba got a [builds page](https://mgba.io/builds/1/).    
Using this we can reliably get a build and avoid the hash mismatch issues.


Calinou (approved):
Thanks!

PR #205, opened 2021-02-21
Update ryujinx-dev.json - simple typo correction

RavenMacDaddy:
Simple typo correction.

PR #207, opened 2021-02-22
Make various improvements to manifests

BlackPowerade:
After looking at mgba and realizing it could use normal versioning after all, I decided to go back through a lot of some older manifests to see if they could be improved, and there are quite a few.    
Rather than do PR spam I will just have a big collection of them here with a summary of the changes.

dolphin(-dev)   
Appdata migration     
new regex in case of new 6.0 stable version

melonds     
Show actual firmware directory in notes     

PCSX2(-dev)
Appdata migration     
improved regex (-dev)     
github based checkver and autoupdate (stable)     

ryujinx-dev
remove unnecessary  `${2}`
fix typo (credit to @RavenMacDaddy #205)

xenia
move to numbered versioning using appveyor builds
persistence
appdata migration


Calinou (changes requested):
It looks like a few packages like `bucket/0ad.json` and `bucket/ark-smart-breeding.json` (and more) were downgraded to their respective previous versions. Could you revert this?

Once this is done, this should be good to merge :slightly_smiling_face:

Calinou (approved):
Thanks a lot :slightly_smiling_face:

PR #212, opened 2021-03-01
openmw-nightly: Fix archive name

apommel:
A new fix for openmw-nightly to fit the artifacts folder structure.


Calinou (approved):
Thanks!

PR #213, opened 2021-03-01
Katrain fix autoupdate

apommel:
The version tags on Github were changed so I had to change the autoupdate URL.


apommel:
Uh there is an additional commit, sorry about that.

PR #214, opened 2021-03-01
katrain: Fix autoupdate

apommel:
The version tags on Github were changed so I had to change the autoupdate URL.


Calinou (approved):
Thanks!

PR #233, opened 2021-04-13
vkquake2: Add version 1.5.7

RavenMacDaddy:
Hopefully this works better for a PR.

EDIT: Changed hash to lower case letters for easier readability.


bucket/vkquake2.json
```diff
@@ -0,0 +1,95 @@
+{
+    "homepage": "https://github.com/kondrak/vkQuake2",
+    "description": "Quake II source port based on v3.21 with Vulkan support.",
```
Calinou, line 3:
```suggestion
    "description": "Quake II source port based on v3.21 with Vulkan support",
```

Calinou (approved):
Thanks!

PR #234, opened 2021-04-14
raze: Update to version 1.0.1

RavenMacDaddy:
- Updated urls to match the new naming scheme; omitting the beta tag.
- This also makes the json update to the latest available version.


Calinou (approved):
Thanks!

PR #248, opened 2021-05-02
mameui: Add description, persist folders

RavenMacDaddy:
- Add description
- Persist folders that I believe to be the most appropriate; those that save things like settings, snapshots, nvram and similar, and not folders like roms and similar since these are often specified to custom directories anyway.

Sadly I can't persist the ini files in the root folder since these are created on the first startup of the software.

### DIFF
--- a/bucket/mameui.json
+++ b/bucket/mameui.json
@@ -1,6 +1,6 @@
 {
     "homepage": "http://www.mameui.info/",
-    "description": "MAME is a multi-purpose emulation framework",
+    "description": "GUI frontend for MAME, the multi-purpose emulation framework",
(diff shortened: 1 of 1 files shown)


bucket/mameui.json
```diff
@@ -1,5 +1,6 @@
 {
     "homepage": "http://www.mameui.info/",
+    "description": "MAME is a multi-purpose emulation framework",
```
Calinou, line 3:
```suggestion
    "description": "GUI frontend for MAME, the multi-purpose emulation framework",
```

Calinou (approved):
Thanks!

PR #250, opened 2021-05-05
dolphin: Fix portable mode broken by c63f4af

mgziminsky:
Commit c63f4af broke dolphin portable mode by removing the `post_install` and instead creating `portable.txt` in `$persist_dir` during `pre_install`, which isn't correct.

Removed that line from `pre_install` and restored the removed `post_install`.


BlackPowerade:
It appears I forgot to put `portable.txt` under `persist`          
```
"persist": [         
    "User",            
    "portable.txt"           
]
```
Also works.

Calinou (approved):
Thanks!

PR #257, opened 2021-05-19
Improve amidst.json

jcwillox:
Take 2 of #184, it keeps the `jar` version of amidst and so should work fine for everyone.
* adds amidst icon to start menu
* shows log output in batch window
* provides access to the amidst cli

I also formatted the file first with `formatjson.ps1`.


Calinou (approved):
Tested locally, it works.

PR #260, opened 2021-05-29
assetstudio: Update to version 0.15.47, add autoupdate

eai04191:
They started using the github release again for the release, so autoupdate is now available.


Calinou (approved):
Thanks!

Issue #259, opened 2021-05-29
openmw-nightly@20210527-9dfba37c: hash check failed


avarun42:
Installing 'openmw-nightly' (20210527-9dfba37c) [64bit]
dl.7z (40.9 MB) [=============================================================================================] 100%
Checking hash of dl.7z ... ERROR Hash check failed!
App:         games/openmw-nightly
URL:         https://gitlab.com/OpenMW/openmw/-/jobs/artifacts/9dfba37c/download?job=Windows_Ninja_Engine_Release#/dl.7z
First bytes: 50 4B 03 04 14 00 08 00
Expected:    50e7282bcaf37c1e8e63e92f85992c6976823a93a25aeaedf86404df53997561
Actual:      f83f9f7733c6f24684fd6628d3b1f9e974b482d5b494fbd3a47a233a39a19410

apommel:
I have seen this too. It seems like the Gitlab artifacts of OpenMW can change after they've been uploaded, even though the version is the same.
Sometimes the hash will be the same, sometimes it won't, but if you skip hash check you will end up with the correct version so it is not a url problem. I will try to take a more thorough look when I have some time.

apommel:
I think I got it. There now are scheduled CI builds everyday, and so scoop downloads the latest build for the relevant commit, and they somehow have a different hash so it does not work. I will look at it to fix it, but it does not seem easy.

Calinou:
Fixed by https://github.com/Calinou/scoop-games/pull/263.

PR #263, opened 2021-06-06
openmw-nightly: Fix autoupdate

apommel:
Should correct #259.
I had to change the entire autoupdate to work on jobs and not commit. It should work, but I will check in some days if there are no issues.


Calinou (approved):
Thanks!

PR #264, opened 2021-06-17
openmw-nightly: Update only if the build passed

apommel:
Since my last fix, the autoupdate could trigger on failed build, which made the installation fail. I updated the regex to check for this.


Calinou (approved):
Thanks!

PR #268, opened 2021-06-20
yquake2: Update to version 8.00

### LINKED ISSUE #266: yquake2 is not updating
So v8.00 has been released, but I can't for the life of me understand how to update the regex code in the current manifest, which at the moment is using the deprecated re function.

BlackPowerade:
Took another shot at it, turns out scoop has a built-in last match function. No need for regex non-sense.
This also updates yquake2 to 8.00
Closes #266


Calinou (approved):
Thanks!

PR #269, opened 2021-06-20
mame: Overhaul

### LINKED ISSUE #244: mame:  version 0.218 and newer are 64-bit only
As of MAME 0.218, [32-bit Windows binaries are no longer supplied](https://www.mamedev.org/?p=479). The binary for the 64-bit version has also been renamed to simply "mame.exe".

I'm not sure if there's a way to limit versions by architecture or if it'd be better to move 32-bit versions to a legacy manifest.

BlackPowerade:
This PR 
1. updates MAME to 0.232
2. Fixes persistence issues caused by scoop trying to persist before `-createconfig` finishes running
3. Prompts 32bit users to install 0.217, by running `scoop install` on an old version of the manifest
4. Removes auto update for 32bit
5. Closes #244

### DIFF
--- a/bucket/mame.json
+++ b/bucket/mame.json
@@ -1,36 +1,49 @@
 {
     "homepage": "http://mamedev.org",
     "description": "Arcade machine emulator",
-    "version": "0.217",
+    "version": "0.232",
     "license": "GPL-2.0-or-later",
     "architecture": {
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #272, opened 2021-06-25
pacmc: Add version 0.1.1

bluefireoly:


Calinou (approved):
Thanks!

bucket/pacmc.json
```diff
@@ -0,0 +1,22 @@
+{
+    "version": "0.1.1",
+    "description": "An easy to use package manager for Fabric Minecraft mods.",
+    "license": "GPL-3.0",
```
Calinou, line 4: Please specify in the project sources whether the license is "GPLv3 only" or "GPLv3 or later". I'll merge this as it is for now, but `GPL-3.0` is not considered a valid SPDX license identifier anymore because of its ambiguity.

bluefireoly: Thanks! I will change it in another PR soon.

PR #284, opened 2021-07-03
retroarch: Add version 1.9.5

NoraTheGamer:


BlackPowerade:
Retroarch is already present in the extras bucket
https://github.com/lukesampson/scoop-extras/blob/master/bucket/retroarch.json

PR #298, opened 2021-08-08
openmw-nightly: Fix checkver regex

apommel:
There seems to have been a change in the GitLab webpage so the regex had to be updated. Hopefully they don't change it too often.


Calinou (approved):
Thanks!

PR #305, opened 2021-09-03
melonds: Update to version 0.9.3, fix autoupdate

RavenMacDaddy:
They added an underscore before 'x64' in the URL making autoupdate string invalid - hopefully this is the only thing they changed this release.

EDIT: check fail says this:

"The following 1 lines contain trailing whitespace: 

C:\projects\scoop-games\bucket\opdessertstorm.json, Line: 31"

Don't know where this comes from as it's nothing to do with this manifest.


Calinou (approved):
Thanks!

PR #310, opened 2021-09-09
Add RPCS3 persist files and folders.

mnxn:
There are some files and folders that aren't persisted for RPCS3:

| Folder | Explanation |
| - | - |
| `dev_flash` | Where the firmware is installed to. Persisting this avoids having to reinstall the firmware every updated. |
| `cache` | PPU modules can take quite bit of time to compile on first run of a game, persisting this allows cache from the previous version to work. |
| `config.yml` | The global emulator configuration. |
| `config` | Game-specific emulator configuration and controller configurations. |
| `games.yml` | The games visible in the game list when you start RPCS3. |
| `Icons` | Custom icon location. |
| `patches` | Downloaded game patches. |

To my knowledge, the built-in RPCS3 updater persists these files and folders, so scoop should too.

PR #311, opened 2021-09-10
rpcs3-dev: Create `.yml` files if they don't exist

mnxn:
Overlooked this in the last PR (#310), scoop needs to create empty `config.yml` and `games.yml` persist files since they are not included in the RPCS3 default files.

Currently scoop creates  `config.yml` and `games.yml` as folders, which causes errors in RPCS3.

I used the same method as the cemu install script:

https://github.com/Calinou/scoop-games/blob/c3f650d568873c15a181611da28a750c7f2d5982/bucket/cemu.json#L20-L22

### DIFF
--- a/bucket/rpcs3-dev.json
+++ b/bucket/rpcs3-dev.json
@@ -24,6 +24,16 @@
             "RPCS3-dev"
         ]
     ],
+    "installer": {
+        "script": [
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #316, opened 2021-09-22
fbneo: Fix autoupdate

RavenMacDaddy:
Noticed that a "v" was missing before the version variable in the autoupdate URL - dunno if it actually mattered, but better safe than sorry.


Calinou (approved):
Thanks!

PR #317, opened 2021-09-22
flycast: Add version 1.1

RavenMacDaddy:
Here's a manifest for the great emulator that's mainly for Sega Dreamcast, but also other platforms like Naomi and Atomiswave (also based on Sega's hardware).

What's missing is a link to its cfg file since this is created on first run - by the way, should files that are linked use a symbolic link or a hard link, and what's the reasoning behind that?

### DIFF
--- a/bucket/flycast.json
+++ b/bucket/flycast.json
@@ -1,0 +1,23 @@
+{
+    "homepage": "https://github.com/flyinghead/flycast",
+    "description": "Flycast is a multi-platform Sega Dreamcast, Naomi and Atomiswave emulator derived from reicast",
+    "version": "1.1",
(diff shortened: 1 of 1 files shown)


Calinou:
> by the way, should files that are linked use a symbolic link or a hard link, and what's the reasoning behind that?

I don't know what kind of link should be used. I guess the most reliable choice should be picked. Note that symbolic links can only be created by non-administrator users on Windows 10; I'm not sure if this limitation applies to hard links as well.

Calinou (approved):
Thanks!

RavenMacDaddy:
Yeah that's fine, as it's now it's also a non-issue since we don't have to make that choice; the manifest can't persist the emu.cfg as it doesn't exist from the get go anyway - and maybe the persist function even makes that choice for us.

PR #318, opened 2021-09-26
pcsx2-dev: fix extract_dir

mnxn:
The last 5 versions of pcsx2-dev have failed to install with an error like this:

```
Extracting dl.7z ... Exception: C:\dev\scoop\apps\scoop\current\lib\core.ps1:510
Line |
 510 |          throw "Could not find '$(fname $from)'! (error $($proc.ExitCo …
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Could not find 'pcsx2-v1.7.0-dev-1806-gc1fea5bc1-windows-x86'! (error 16)
```

This is because the directory inside the 7z file seems to no longer include the build ID or commit hash.

This PR removes the build ID and commit hash from the current `extract_dir` as well as the template in `autoupdate`.


mnxn:
Nevermind. It seems the problem is that its downloading the wrong file completely.
The metadata in the 7z file shows that the files are from 2020-05-07.

Calinou:
Superseded by https://github.com/Calinou/scoop-games/pull/319.

PR #366, opened 2021-10-30
Update angband.json

satorope:
- Supports autoupdate
- Add various Persist data
- Make portable
- Other repairs

I made various changes, so I will update them all at once.
The hash of the file is obtained by autoupdate.

Since Angband and its modifications can't persist INI files, I've made sure that updates and so on are backed up.

### DIFF
--- a/bucket/angband.json
+++ b/bucket/angband.json
@@ -1,27 +1,44 @@
 {
-    "version": "4.2.0",
-    "description": "A free single-player dungeon exploration game",
+    "version": "4.2.3",
+    "description": "A free single-player dungeon exploration Roguelike",
     "homepage": "https://rephial.org/",
     "license": "GPL-2.0",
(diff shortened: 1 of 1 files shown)


bucket/angband.json
```diff
@@ -1,27 +1,44 @@
...
+        "}"
+    ],
+    "uninstaller": {
+        "script": [
+            "'angband.INI' | ForEach-Object {",
+            "    Copy-Item \"$dir\\$_\" \"$persist_dir\\$_.bak\" -Force",
+            "}"
+        ]
+    },
+    "notes": "Configuration files cannot be persisted, but will be retained during the update"
```
Calinou, line 43:
```suggestion
    "notes": "Configuration files cannot be persisted, but will be retained during package updates."
```

Calinou (approved):
Thanks!

PR #376, opened 2021-10-30
cataclysm-dda: Add persist, rename shortcuts

satorope:
Autoupdate now works.
Add persist data.
Rename shortcuts according to official site.

### DIFF
--- a/bucket/cataclysm-dda.json
+++ b/bucket/cataclysm-dda.json
@@ -2,10 +2,7 @@
     "homepage": "https://cataclysmdda.org",
     "description": "Roguelike in a post-apocalyptic world (with text-based graphics)",
     "version": "0.D",
-    "license": {
-        "identifier": "CC-BY-SA-3.0,GPL-2.0+,OFL-1.0,BSL-1.0,Zlib,MIT,BSD(libbacktrace)",
-        "url": "https://github.com/CleverRaven/Cataclysm-DDA/blob/master/LICENSE.txt"
-    },
+    "license": "CC-BY-SA-3.0",
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


bucket/cataclysm-dda.json
```diff
@@ -1,8 +1,11 @@
 {
     "homepage": "https://cataclysmdda.org",
-    "description": "Turn-based survival game set in a post-apocalyptic world (with text-based graphics)",
+    "description": "Roguelike in a post-apocalyptic world (with text-based graphics)",
     "version": "0.D",
-    "license": "CC-BY-SA-3.0",
+    "license": {
+        "identifier": "CC-BY-SA-3.0,GPL-2.0+,OFL-1.0,BSL-1.0,Zlib,MIT,BSD(libbacktrace)",
+        "url": "https://github.com/CleverRaven/Cataclysm-DDA/blob/master/LICENSE.txt"
+    },
```
Calinou, line 8:
```suggestion
    "license": "CC-BY-SA-3.0",
```

Calinou (approved):
Thanks!

satorope:
CC-BY-SA-3.0 is a license for Cataclysm, which is the basis of CDDA, and I suspect that fonts and other things are not included in that license.
The license text looks the same.
https://github.com/Whales/Cataclysm/blob/master/LICENSE

The creator should define the Cataclysm-dda license, but it seems to be neglecting that and listing the licenses for the materials used.
CC-BY-SA-3.0 is not possible when GPL etc. are incorporated.
In particular, fonts are only attached, so license restrictions will be inherited when redistributed.

I also mistook the OFL license.
OFL-1.1 is correct.

satorope:
If it is a rule to write only the license of the main program itself, I will leave the decision to you because the above story may be inappropriate.

PR #393, opened 2021-10-30
infraarcana: Update to version 20211017-c078746a, add autoupdate

satorope:
Autoupdate now works.
It was changed to the development version according to the manufacturer's recommendation.


Calinou (approved):
Thanks!

PR #400, opened 2021-10-30
melonds: Exclude BIOS and firmware from persist

### LINKED ISSUE #361: melonds@0.9.3: Exclude bios and firmware from persist
Melonds does not load automatically when you place bios9.bin; you must select bios manually.
Melonds creates firmware.bin.bak as a backup of firmware.bin and then edits the contents of firmware.bin.
However, the edited firmware.bin is not reflected in the persist folder, and firmware.bin.bak is unrelated to the persist folder.

To avoid misunderstandings that firmware.bin is stored in the persist folder, I think it is better to remove it.

satorope:
Exclude bios and firmware from persist
Closes #361

### DIFF
--- a/bucket/melonds.json
+++ b/bucket/melonds.json
@@ -7,21 +7,6 @@
     "hash": "108693998ce58175b4acb02788177f80e9e277658fa2cc9c62a2c7ae34181b10",
     "installer": {
         "script": [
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

Issue #334, opened 2021-10-15
Some game versions were old


satorope:
Some game versions were old, so I will report the latest version.
Please forgive me for cutting corners.

freedoom 0.12.1
freedroidrpg 1.0rc2 
ivan 058
odamex 0.9.5
openttd 1.11.2
paintball2 2.0 alpha build 43
prboom-plus 2.6.1um
ps3-system-software 4.8.8
pychess 1.0.3 
rbdoom-3-bfg 1.2.0 
simutrans 122.0
spearmint 1.0.3
speed-dreams 2.2.3
wyrmsun 5.2.2
xmoto 0.6.1

PR #413, opened 2021-10-30
Update openttd.json

satorope:
Autoupdate now works.
Make portable.
Autoupdate can manage only one file.
Graphics and music data are managed by a different version and should be managed by the built-in downloader.

Closes #334

### DIFF
--- a/bucket/openttd.json
+++ b/bucket/openttd.json
@@ -61,5 +61,5 @@
             "find": "checksum-$basename[\\s\\S]*?([a-fA-F\\d]{64})"
         }
     },
-    "notes": "To save the data in persist folder, run OpenTTD from the Start menu or openttd-persist.exe on CLI"
+    "notes": "To save the data in persist folder, run OpenTTD from the Start menu or openttd.exe on CLI"
 }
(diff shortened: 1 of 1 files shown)


bucket/openttd.json
```diff
@@ -1,48 +1,65 @@
...
+        "content_download",
+        "newgrf",
+        "save",
+        "scenario",
+        "screenshot"
+    ],
+    "bin": [
+        [
+            "openttd.exe",
+            "openttd-persist",
```
Calinou, line 33: Why not just call it `openttd` instead of `openttd-persist`? The shortcut entry below isn't added to the `PATH` automatically.

satorope: It was the manifest I started with, so I wasn't sure if I could add an option and use the same name.
Use the same name as other manifests.

satorope:
4cb4957
Restored the original name.
ec5f240
Notes has been modified.

ThatLinuxNerd:
Maybe they can be excluded? The game prompts you to download open-gfx on first launch, while openmsx and opensfx are generally installed afterwards by the player.

Calinou:
Superseded by https://github.com/Calinou/scoop-games/pull/830. Thanks for the contribution nonetheless!

PR #442, opened 2021-10-30
Update wyrmsun.json

satorope:
Autoupdate now works.
Since AutoUpdate can only handle one file, separate exe.
Closes #334


satorope:
309ef21
Changed to direct download of the same version of exe file
56a76db
Built-in dl now shows progress when downloading EXE files with matching versions

PR #448, opened 2021-10-30
Create wyrmsun-exe.json

satorope:
wyrmsun's exe file

### DIFF
--- a/bucket/wyrmsun-exe.json
+++ b/bucket/wyrmsun-exe.json
@@ -1,0 +1,34 @@
+{
+    "homepage": "https://andrettin.github.io/",
+    "description": "Binary for the Wyrmsun on Windows",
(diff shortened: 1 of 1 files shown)


Calinou:
The use case for this package seems a bit too specialized for me, considering we already have https://github.com/Calinou/scoop-games/blob/master/bucket/wyrmsun.json. I'd rather only have a single package that works out of the box. For more advanced use cases, you can perform the installation manually.

satorope:
If more than one file is updated by version, autoupdate cannot handle it and you have to update the manifest manually.
Also, Wyrmsun is a Steam distribution game that has been updated to some extent.
It's pending until I come up with a solution.

satorope:
I changed the method to download the same version of exe file directly.

PR #451, opened 2021-10-30
Create broguece.json

satorope:
I overlooked the following. Why not add it to main?
https://github.com/Calinou/scoop-games/blob/3e25970af514f7bd69445d88f1a326bc497f9b96/bucket/broguece.json

After formatting with formatjson.ps1, I have fixed the autoupdate and comment line count.

### DIFF
--- a/bucket/broguece.json
+++ b/bucket/broguece.json
@@ -1,0 +1,44 @@
+{
+    "version": "1.10.1",
+    "description": "A community-lead fork of the much-loved minimalist roguelike game, Brogue",
+    "homepage": "https://github.com/tmewett/BrogueCE",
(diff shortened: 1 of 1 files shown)


bucket/broguece.json
```diff
@@ -0,0 +1,44 @@
+{
+    "version": "1.10.1",
+    "description": "A community-lead fork of the much-loved minimalist roguelike game, Brogue",
+    "homepage": "https://github.com/tmewett/BrogueCE",
+    "license": "AGPL-3.0",
```
Calinou, line 5:
```suggestion
    "license": "AGPL-3.0-or-later",
```

PR #455, opened 2021-10-31
ares: Add version 124

### LINKED ISSUE #452: Add ares
Taking advantage of # 300's request, I would like to add emulators that are highly rated on reddit but are not built into RetroArch.
Excluding redream, which is a challenge due to the troublesome handling of Persist.

ares
https://ares-emulator.github.io/

satorope:
Configure as Portable Version

Closes #452


Calinou (approved):
Thanks!

PR #458, opened 2021-10-31
nethack: Show data paths after installation, fix autoupdate

satorope:
Autoupdate now works.
After installation, a list of folders to store data is displayed.
It could theoretically be portable, but it's still a challenge.


Calinou (approved):
Thanks!

PR #474, opened 2021-11-09
Update freeorion-test.json

satorope:
Like the official installer, I have added a full-screen shortcut.
It was mentioned in the official wiki, so I decided it was necessary.
The bin was removed because it could not be executed normally outside of the installation folder.


Calinou:
This needs a rebase before it can be merged.

PR #475, opened 2021-11-09
blade-of-agony: Update to version 3.1, use ModDB download, add persist

satorope:
Upgraded to the final version.
Added a script to download from a URL randomly generated by moddb.com.
Added Persist.


Calinou (approved):
Thanks!

PR #480, opened 2021-11-09
q2pro: Add 64-bit and persist, use versioned release

satorope:
The version is indicated.
Added 64 bit version.
Automatic update enabled.
Libraries in the Persist folder are now updated on update.
The menu file is assumed to be customized, so it is excluded.


Calinou (approved):
Thanks!

PR #481, opened 2021-11-09
Update yquake2.json

satorope:
Added Persist such as music which is mentioned in the notebook.
Persist libraries are updated during automatic updates.


Calinou (approved):
Thanks!

Calinou:
This needs a rebase before it can be merged.

beyondmeat:
Also the persist stuff can be done a better way than what's done here.

PR #488, opened 2021-12-18
flycast: Update to version 1.2

RavenMacDaddy:
Updates the manifest to v1.2.

Also updates the autoupdate section since the words have been thrown around and skips the V before the version.


RavenMacDaddy:
Next commit also fixes the pre-install by adding curly braces around the action after the condition.

Calinou (approved):
Thanks!

PR #489, opened 2021-12-18
mgba: Update to version 0.9.3, tweak checkver and autoupdate

RavenMacDaddy:
mGBA has had very simple versioning for a while.

This updates:

- The version to 0.9.3 which was released almost a day ago
- Checkver to simply use the github method
- Autoupdate to simply use the $version for matching


Calinou (approved):
Thanks!

Issue #315, opened 2021-09-19
DOOM Retro doesn't create a cfg file making the persist broken


RavenMacDaddy:
When installing DOOM Retro, it doesn't create a doomretro.cfg in its app folder - or rather, it's supposed to create it at the first execution, however this still breaks the persist process as it will instead create a folder called doomretro.cfg.

If I do however delete the persisted folder, start the game so that the file called doomretro.cfg is created, and finally reset the app, it's persisted as intended.

Question is if there's some way to create this file with Scoop before starting the persist process in order for the manifest as a whole to work as intended already on the initial install?

EDIT: Checking my older issue regarding this, I'm not sure how to completely fix it since keybindings will be missing if you manually create the file yourself.

It would be one thing if a standard cfg file was somehow bundled in the manifest, though I'm not sure if the app will populate it with eventual new or changed features or values present in future versions in that case.

Somehow the part of the app that creates the cfg file must be triggered before the persist process for it all to work.

EDIT 2: Now that I try to reset the app once the proper cfg file is in place (one that I edited according to my preferences), it will rename it with ".original" at the end and instead create a folder named the same once again.

Cheers

RavenMacDaddy:
Closing this for now.

Hopefully this gets sorted out with time as new versions are released for the app, rather than us trying to work around its config issues.

PR #490, opened 2021-12-19
doom-retro: Fix config file persist and add info to notes

RavenMacDaddy:
This fixes the persist of the config file, though it doesn't fix issue #315 since the controls still end up empty.

Therefore I've added this information in the notes to simply improve the situation a bit.


RavenMacDaddy:
Last commit fixes white space error.

Calinou (approved):
Thanks!

RavenMacDaddy:
Fixes punctuation... Yeah I'm a bit of a perfectionist when it comes to my own work.

PR #491, opened 2021-12-19
rbxfpsunlocker: Update to version 4.4.1, fix autoupdate

LewisTehMinerz:
changes checkver to `github` mode (as it should've been when this manifest was added, rbxfpsunlocker uses the releases system on github so we can benefit from it here) and changes the autoupdate urls to use github release urls instead of file IDs.

### DIFF
--- a/bucket/rbxfpsunlocker.json
+++ b/bucket/rbxfpsunlocker.json
@@ -1,12 +1,12 @@
 {
-    "version": "4.4.0",
+    "version": "4.4.1",
     "description": "Removes Roblox's default 60 FPS cap",
     "homepage": "https://github.com/axstin/rbxfpsunlocker",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


Calinou (approved):

LewisTehMinerz:
glad to contribute :)

sitiom:
> (as it should've been when this manifest was added, rbxfpsunlocker uses the releases system on github so we can benefit from it here) and changes the autoupdate urls to use github release urls instead of file IDs.

@LewisTehMinerz When I added the manifest, rbxfpsunlocker wasn't using the release assets and instead uploaded it directly to the release description (which explains why the URL is a file ID), and so you could not use the `github` checkver shorthand.

LewisTehMinerz:
good spot, didn't see that. ignore me then!

PR #495, opened 2021-12-22
dxvk: Add version 1.9.2

RavenMacDaddy:
A Vulkan-based translation layer for Direct3D 9/10/11.

### DIFF
--- a/bucket/dxvk.json
+++ b/bucket/dxvk.json
@@ -1,6 +1,6 @@
 {
   "homepage": "https://github.com/doitsujin/dxvk",
-  "description": "A Vulkan-based translation layer for Direct3D 9/10/11.",
+  "description": "A Vulkan-based translation layer for Direct3D 9/10/11",
   "version": "1.9.2",
   "license": "Zlib",
   "url": "https://github.com/doitsujin/dxvk/releases/download/v1.9.2/dxvk-1.9.2.tar.gz",
--- a/bucket/dxvk.json
+++ b/bucket/dxvk.json
@@ -1,0 +1,23 @@
+{
+  "homepage": "https://github.com/doitsujin/dxvk",
+  "description": "A Vulkan-based translation layer for Direct3D 9/10/11.",
+  "version": "1.9.2",
+  "license": "Zlib",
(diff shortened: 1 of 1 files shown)


bucket/dxvk.json
```diff
@@ -0,0 +1,23 @@
+{
+  "homepage": "https://github.com/doitsujin/dxvk",
+  "description": "A Vulkan-based translation layer for Direct3D 9/10/11.",
```
Calinou, line 3:
```suggestion
  "description": "A Vulkan-based translation layer for Direct3D 9/10/11",
```

bucket/dxvk.json
```diff
@@ -0,0 +1,23 @@
...
+  "notes": [
+    "",
+    "For installation, copy the corresponding DLL file depending on the architecture (e.g. x32 or x64) and Direct3D version (e.g. D3D9, 10 or 11) that's used by the app, to the directory where the app loads them from.",
+    "",
+    "In most cases, this is the same folder as where the executable (.exe file) is stored, though some games, for example Source based ones, uses a sub-folder.",
+    "",
+    "In the example given, the sub-folder named 'bin' is used.",
+    "",
+    "Finding other DLL files is usually a good indication of where to copy the DLL file from DXVK.",
+    ""
```
Calinou, line 21:
```suggestion
    "",
    "For installation, copy the corresponding DLL file depending on the architecture (e.g. x32 or x64) and Direct3D version (e.g. D3D9, 10 or 11) that's used by the app, to the directory where the app loads them from.",
    "",
    "In most cases, this is the same folder as where the executable (.exe file) is stored, though some games (such as Source-based games) use a sub-folder instead.",
    "",
    "In the example given, the sub-folder named 'bin' is used.",
    "",
    "Finding other DLL files is usually a good indication of where to copy the DLL file from DXVK.",
    ""
```

bucket/dxvk.json
Calinou, line 3:
```suggestion
  "description": "A Vulkan-based translation layer for Direct3D 9/10/11",
```

PR #496, opened 2021-12-22
nanoboyadvance: Fix hash

RavenMacDaddy:
Seems like the download has gotten a new hash making the download fail in Scoop, this corrects it to the latest one.


Calinou (approved):
Thanks!

PR #505, opened 2022-01-10
dosbox-staging: Add version 0.78.1

RavenMacDaddy:
DOSBox Staging is DOS/x86 emulator focusing on ease of use. Based on DOSBox.

---

Tested first by running it separately outside of Scoop to figure out persist/portable mode.

Basically if you don't have a dosbox-staging.conf in your root, it will use Local AppData - whilst if you do have it, it will use the config in the root, and create folders for glshaders, mt32-roms, and soundfonts. So all of this is persisted in the manifest and the config file is created through pre install with an If statement.


RavenMacDaddy:
Already a part of the Extras bucket, closing.

PR #516, opened 2022-01-28
scummvm: Add version 2.5.1

RavenMacDaddy:
Credit goes to [hermanjustnu](https://github.com/hermanjustnu/scoop-emulators/blob/master/bucket/scummvm.json).

The manifest looks good out the gate.


Calinou (approved):
Thanks!

PR #519, opened 2022-01-30
doom-retro: Improve the Notes section

RavenMacDaddy:
The notes described the _doom dir as being under the DOOM Retro persist dir, which is not correct - so this corrects that.

It also inserts some space in the notes section to make it more readable and aesthetically pleasing.

### DIFF
--- a/bucket/doom-retro.json
+++ b/bucket/doom-retro.json
@@ -47,8 +47,10 @@
         }
     },
     "notes": [
+        "",
         "ATTENTION: DOOM Retro requires WAD files, e.g. from a commercial DOOM copy (DOOM 1, 2, Ultimate DOOM, etc).",
-        "Place WAD files in the _doom directory under: $persist_dir",
+        "",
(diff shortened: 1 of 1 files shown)


bucket/doom-retro.json
```diff
@@ -47,17 +47,20 @@
...
+        "Place the WAD files in the _doom directory which is under your persist dir.",
         "",
         "If you want to bind controls in the config file, here are some examples:",
         "",
         "bind capslock +alwaysrun",
         "bind 's' +back",
         "bind mouse1 +fire",
         "",
-        "Basically it's: bind control +action",
+        "Basically it's: bind control +action, where singular letters and signs need to have apostrophes surrounding them in the config file.",
```
Calinou, line 61:
```suggestion
        "Basically, it's: bind control +action, where singular letters and signs need to have apostrophes surrounding them in the config file.",
```

RavenMacDaddy:
Looks good to me.

PR #520, opened 2022-01-31
xenia: Fix persist list

mnxn:
The persist list was incorrectly nested, which caused incorrect persisting after an update.

Specifically, it was causing the contents of all the files to be written to `portable.txt` and skipping the other files:

image


Calinou (approved):
Thanks!

PR #521, opened 2022-01-31
Remove MCServMan project (deleted and discontinued)

CrazyWillBear:


Calinou (approved):
Thanks!

PS: In the future, please consider *archiving* your repository instead of removing it altogether. This helps preserve software that may still be useful to others :slightly_smiling_face:

PR #535, opened 2022-02-12
retroarch-nightly: Add version nightly

beyondmeat:


bucket/retroarch-nightly.json
```diff
@@ -0,0 +1,69 @@
...
+        "screenshots",
+        "states",
+        "system",
+        "thumbnails",
+        "retroarch.cfg"
+    ],
+    "checkver": {
+        "url": "http://worldtimeapi.org/api/timezone/America/Los_Angeles",
+        "regex": "utc_datetime\":\"(\\d{4}-\\d{2}-\\d{2})"
+    },
```
Calinou, line 53: Since this won't be 100% in sync with RetorArch nightly updates, I think it's better to drop the `hash` and `checkver` entries and set the version identifier to `nightly`. This skips hash verification entirely and prevents installations from failing due to mismatching hashes.

beyondmeat:
@Calinou checkver doesn't seem to run now with version: nightly, just blank/null response

Calinou:
> @Calinou checkver doesn't seem to run now with version: nightly, just blank/null response

Indeed, that's how it's designed to work :slightly_smiling_face:

PR #539, opened 2022-02-13
rpcs3: Suggest vcredist2019 instead of depending on it

beyondmeat:
fixes install when you don't have extras bucket added


Calinou (approved):
Thanks!

PR #545, opened 2022-02-13
Normalized shortcuts for various apps

beyondmeat:


mnxn:
Wouldn't this prevent users from having multiple versions of a program installed?

Installing both `dolphin-dev` and `dolphin-beta` currently creates `Dolphin (Development)` and `Dolphin (Beta)` shortcuts. It looks like with this change there would only be one shortcut?

I would prefer to keep the shortcut names the way they are now.

Calinou:
I also agree that shortcut names for different app versions should be kept as-is. It's important to be able to use both stable and beta releases in parallel for comparison and bug reporting purposes :slightly_smiling_face:

beyondmeat:
There are apps that have nightly or dev versions alongside stable that do not differ in shortcuts. If I recall correctly @Calinou you had me change the Retroarch nighty shortcut to just be "Retroarch" for simplicity, so I assumed that was the preferred way.

beyondmeat:
@Calinou would you mind going through the alt versions of what it's the repo and updating them to fit that standard? I think some need fixed, but you have a better idea of how you want it.

Calinou:
> @Calinou would you mind going through the alt versions of what is in the repo and updating them to fit that standard? I think some need fixed, you have a better idea.

I'm pretty busy with things at the moment, so I unfortunately don't have the time to do that just yet.

PR #550, opened 2022-02-17
ppsspp-dev: Fix download link to be downloaded as 7z file

kiennq:
The current link will be downloaded as .PHP file


Calinou (approved):
Thanks!

PR #551, opened 2022-02-17
ci: Automatically try to update packages

kiennq:


Calinou:
@r15ch13 still has their autoupdater running on this repository, should he stop it before I merge this pR?

beyondmeat:
@Calinou if this works, this would be better and update more often

r15ch13:
Should I deactivate it on my end?

Calinou:
> Should I deactivate it on my end?

Yes, I would recommend disabling your updater if this new updater turns out to work out nicely.

PR #555, opened 2022-02-18
coinkiller: Add version 0.15

furtabs:


UnicorNora:
Merge?

beyondmeat:
Need to sort the manifest.

I recommend using vs code and open the workspace file, it will recommend the extensions needed. Then just right click and sort or save the file to resort it to the spec.

PR #566, opened 2022-02-27
advanced-genie-editor: Remove as it's broken

beyondmeat:
Creator does not wish for this to be added to scoop (see project readme). This has been broken for quite some time now.


Calinou (approved):
Thanks!

PR #567, opened 2022-02-27
brogue: Remove as it's broken

beyondmeat:
Been broken for a while. This has been deprecated and replaced with another fork (BrogueCE). See google site readme or
https://github.com/tmewett/BrogueCE

Can add BrougeCE in another PR if users want it.


Calinou (approved):
Thanks!

Issue #303, opened 2021-08-28
ioquake3 error


naviabbot:
Upon attempting to install ioquake3, the following error message is displayed:

```
The remote server returned an error: (404) Not Found.
URL https://ioquake3.org/files/jenkins/latest/gcc/no_options/release-mingw32-x86_64.zip is not valid
```

Upon investigation, it appears that the correct address should be `https://files.ioquake3.org/Windows.zip`.

Calinou:
https://files.ioquake3.org/Windows.zip contains a ZIP archive within a ZIP archive, which means it's more difficult to support within Scoop. I asked the maintainer to change it to a single, non-nested ZIP archive.

PR #569, opened 2022-02-28
ioquake3: Remove as it's broken

beyondmeat:
The new download location points to a zip file within a zip file, which would need extracted. So Windows.zip contains release-mingw64-x86_64.zip, which holds the ioquake3 files. AFAIK scoop can't extract a zip inside another zip.

*Maintainer edit: This closes https://github.com/Calinou/scoop-games/issues/303.*


Calinou:
Thanks!

As an alternative, `spearmint` can be used (and `quake3e` could be added to this bucket too).

PR #576, opened 2022-03-02
polymc: Fix hash

lem6ns:


Calinou (approved):
Thanks!

Calinou:
PS: In the future, remember not to update release files once they were published. Instead, _always_ publish a new release (even if the changes are very minor). This avoids breaking hashes for existing releases. Also, please don't remove old releases entirely as they're useful to keep for archival purposes.

PR #577, opened 2022-03-03
ares: Fix autoupdate and extract_dir

RavenMacDaddy:
This should fix the extract dir by correcting it to the current version which is 127, and autoupdate by using cleanVersion.


Calinou (approved):
Thanks!

PR #582, opened 2022-03-06
appveyor: update scoop repo url, update poweshell 6 to current

beyondmeat:
reduce git depth.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-version: '{build}-{branch}'
+version: "{build}-{branch}"
 
 branches:
   except:
@@ -6,7 +6,7 @@ branches:
 
 build: off
 deploy: off
-clone_depth: 49
+clone_depth: 2
 
 image: Visual Studio 2019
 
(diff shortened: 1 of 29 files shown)


appveyor.yml
```diff
@@ -18,17 +18,16 @@ environment:
   innounp: '%scoop_helpers%\innounp\innounp.exe'
   matrix:
     - PowerShell: 5
-    - PowerShell: 6
+    - PowerShell: Current
```
Calinou, line 21: I'd prefer if we list the major version manually. This prevents the CI from suddenly breaking when upstream releases an incompatible version.

Calinou (changes requested):

appveyor.yml
```diff
@@ -39,7 +38,7 @@ for:
     - ps: . "$env:SCOOP_HOME\test\bin\test.ps1" -TestPath "$env:APPVEYOR_BUILD_FOLDER"
 - matrix:
     only:
-      - PowerShell: 6
+        - PowerShell: Current
```
Calinou, line 41: Same as above.

beyondmeat: I prefer to know as soon as a major version breaks and not rely on user reports, but I have made the change you requested.

beyondmeat:
@Calinou  ready

PR #580, opened 2022-03-05
Sort and format manifests

beyondmeat:
No changes to manifests besides cosmetic.
closes #579

PR #584, opened 2022-03-06
fbneo: Adjust the manifest order and Notes

RavenMacDaddy:
- Adjusts the manifest order according to the [contribution guidelines](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).
- Generalizes the contents of notes, which has the upside of making the sentences themselves read better as well.

### DIFF
--- a/bucket/fbneo.json
+++ b/bucket/fbneo.json
@@ -1,23 +1,25 @@
 {
-    "homepage": "https://github.com/finalburnneo/FBNeo",
-    "description": "Emulator for arcade games and select consoles",
     "version": "1.0.0.2",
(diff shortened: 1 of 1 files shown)


Calinou:
The suggested sorting order was changed here: https://github.com/Calinou/scoop-games/pull/580

@beyondmeat Should it be changed to match the upstream Scoop contribution guidelines?

beyondmeat:
@Calinou I was unaware they already had a preferred sort. You can merge this in and then I can adjust the sort and make a PR to adjust everything to that

PR #601, opened 2022-03-28
PolyMC: Update to version 1.1.1, fix autoupdate, remove useless `persist`

DioEgizio:


Calinou:
Are you sure the `persist` entry and preinstall script are no longer needed? If these no longer work, users will lose their data when updating PolyMC.

DioEgizio:
> Are you sure the `persist` entry and preinstall script are no longer needed? If these no longer work, users will lose their data when updating PolyMC.
> 
> Also, `autoupdate` could be modified to support 64-bit downloads. This can be done in a future PR as it's not critical.

Since 1.1.0, PolyMC stores stuff in APPDATA, (for a bug).
On 1.1.1 there are both a portable and a notportable build. Since this package currently uses the notportable build, it's completely useless to keep the persists

Calinou (approved):
Thanks!

PR #604, opened 2022-04-06
pcsx2-dev: Remove 32-bit builds

mnxn:
The PCSX2 repository no longer provides 32 bit builds in their releases page: https://github.com/PCSX2/pcsx2/commit/8a4f1ef51a037d77d2746304d89bef6a325a790b

I believe the now broken `32bit` section in this manifest was preventing it from being autoupdated.


Calinou (approved):
Thanks!

PR #620, opened 2022-05-11
l4d2modmanager: Fix persist with a pre_install script

ClassicDarkChocolate:
Fix persist with a pre_install script.

> - [x] stable
> - [ ] beta
> - [ ] dev/nightly/canary


Calinou (approved):
Thanks!

Issue #612, opened 2022-04-17
Cannot autoupdate to MAME 0.242, and more packages


heroku-miraheze:
https://github.com/Calinou/scoop-games/actions/runs/2179935812

PR #624, opened 2022-05-19
mame: Update to version 0.243, fix autoupdate

makamys:
The dummy URL added in https://github.com/Calinou/scoop-games/pull/269 for the 32-bit version of MAME is no longer online. This caused MAME's autoupdate to stop working, as reported in #612.

This PR changes it to the URL of the manifest JSON itself, which should stay valid for a long time. I also updated the manifest using `check-ver.ps1`.

### DIFF
--- a/bucket/mame.json
+++ b/bucket/mame.json
@@ -1,5 +1,5 @@
 {
-    "version": "0.241",
+    "version": "0.243",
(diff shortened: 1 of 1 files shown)


Calinou (approved):

PR #625, opened 2022-05-22
rpcs3: Fix `config.yml` and `games.yml` not persisting between upgrades

mundifex:
RPCS3 overwrites files config.yml and games.yml when saving changes to them, which eliminates the hardlinks created by Scoop when installing the app. Without the hardlinks, these files do not persist when un/re-installing the app.

This PR works around the overwrite behavior by copying the .yml files to the persist directory before uninstall (`"uninstaller"` definition) and copying them back to the app during installation (`"pre_install"` definition). Because Scoop's default persist handling does not work for these files, I also removed them from the `"persist"` definition.


Calinou (approved):

PR #630, opened 2022-05-28
melonds: Update to 0.9.4

### LINKED ISSUE #617: melonDS Outdated
melonDS.json is outdated. The manifest has 0.9.3 and the latest version is 0.9.4.

BlackPowerade:
Closes #617 
The url scheme changed

### DIFF
--- a/bucket/melonds.json
+++ b/bucket/melonds.json
@@ -6,10 +6,6 @@
         "identifier": "GPL-3.0",
         "url": "https://github.com/Arisotura/melonDS/blob/master/LICENSE"
     },
-    "notes": [
-        "ATTENTION: MelonDS requires BIOS and firmware files dumped from a Nintendo DS to function.",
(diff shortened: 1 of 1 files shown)


bucket/melonds.json
```diff
@@ -10,8 +10,8 @@
         "ATTENTION: MelonDS requires BIOS and firmware files dumped from a Nintendo DS to function.",
         "Visit http://melonds.kuribo64.net/faq.php for more information."
```
Calinou, line 11: According to the release notes, a BIOS file is no longer an absolute requirement for DS mode operation (although I assume it works better with a BIOS). It's only required for DSi mode operation.

BlackPowerade: Neat, didn't know that changed

Calinou (approved):
Thanks!

PR #633, opened 2022-05-29
rpcs3: Fix persistence and disable built-in autoupdate

BlackPowerade:
Old method was essentially doing persistence manually on every install/upgrade


Calinou (approved):
Thanks!

PR #640, opened 2022-06-11
openmw-nightly: Fix checkver

apommel:
A fix for openmw-nightly because the automatic builds were changed. I added a new check in the regex to only update if the commit was on the master branch.


Calinou (approved):
Thanks!

PR #641, opened 2022-06-11
openmw-nightly: Fix install

apommel:
I just realized the install script was not working with my last PR (I could not test it properly because I can't anymore install apps directly from json manifests with scoop for some reason). I tested the commands one by one now so it should work this time.
Sorry about that.


apommel:
I don't know why it displays so much commits as I rebased my branch before, you may want to squash that.

Calinou (approved):
Thanks!

PR #642, opened 2022-06-14
undertale-mod-tool: Fix persist

### LINKED ISSUE #597: undertale-mod-tool: Scoop file permission bug?
"Access Denied"  (only happens when I install via scoop)
Screenshot_20

issaclin32:
* Fixes #597

* `UndertaleModTool.exe.config` does not exist in the archive, causing *Scoop* to persist it as a **folder**, which causes the access error to happen.


Calinou (approved):
Thanks!

PR #649, opened 2022-06-30
(chore): Fix git conflicts with line endings

issaclin32:
This avoids **git conflicts** due to inconsistant line endings (`LF`/`CRLF`), which will sometimes cause users unable to update the bucket.

The fix was applied to all buckets under `ScoopInstaller` org.

related:
https://github.com/ScoopInstaller/Extras/issues/7546
https://github.com/ScoopInstaller/Extras/pull/8285


Calinou (approved):
Thanks!

PR #651, opened 2022-07-03
yquake2-ref-vk: Add version 1.0.2

RavenMacDaddy:
This is the vkQuake2 vulkan renderer library ported to Yamagi Quake II.

> - [x] stable
> - [ ] beta
> - [ ] dev/nightly/canary

### DIFF
--- a/bucket/yquake2-ref-vk.json
+++ b/bucket/yquake2-ref-vk.json
@@ -1,6 +1,6 @@
 {
     "version": "1.0.2",
-    "description": "This is the vkQuake2 vulkan renderer library ported to Yamagi Quake II",
+    "description": "vkQuake2's Vulkan renderer library ported for use with Yamagi Quake II",
     "homepage": "https://github.com/yquake2/ref_vk",
     "license": "https://github.com/yquake2/ref_vk/blob/master/LICENSE",
     "depends": "yquake2",
--- a/bucket/yquake2-ref-vk.json
+++ b/bucket/yquake2-ref-vk.json
@@ -1,0 +1,27 @@
+{
+    "version": "1.0.2",
+    "description": "This is the vkQuake2 vulkan renderer library ported to Yamagi Quake II",
+    "homepage": "https://github.com/yquake2/ref_vk",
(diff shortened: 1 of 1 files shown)


Calinou:
Shouldn't we modify the yquake2 manifest to automatically install the Vulkan renderer instead? I believe it'll still default to OpenGL out of the box, so it'll keep working on GPUs that don't support Vulkan.

RavenMacDaddy:
How do we do that in the original manifest? 

I thought we'd to create this first and then make a `suggest`?

Calinou:
> How do we do that in the original manifest?

I guess we can add a second download URL, similar to [openttd](https://github.com/Calinou/scoop-games/blob/master/bucket/openttd.json).

> I thought we'd to create this first and then make a `suggest`?

That can be done, but since the download size is just a few megabytes, I wouldn't bother.

RavenMacDaddy:
> > How do we do that in the original manifest?
> 
> I guess we can add a second download URL, similar to [openttd](https://github.com/Calinou/scoop-games/blob/master/bucket/openttd.json).
> 
> > I thought we'd to create this first and then make a `suggest`?
> 
> That can be done, but since the download size for the Vulkan renderer ZIP is only 300 KB, I wouldn't bother.

Now I get it, didn't know that possibility - on it.

RavenMacDaddy:
It looks like this needs to be its own manifest still since I can't find a way to make `version`, `extract_dir`, `checkver` and `autoupdate` take care of the vulkan library as well.

Calinou:
Alright, I guess we can merge this as-is then.

Feel free to open a PR to add a `suggest` to yquake2 for yquake2-ref-vk.

bucket/yquake2-ref-vk.json
```diff
@@ -0,0 +1,27 @@
+{
+    "version": "1.0.2",
+    "description": "This is the vkQuake2 vulkan renderer library ported to Yamagi Quake II",
```
Calinou, line 3:
```suggestion
    "description": "vkQuake2's Vulkan renderer library ported for use with Yamagi Quake II",
```

Calinou (approved):
Thanks!

PR #653, opened 2022-07-06
raine: Use HTTPS for autoupdate

RavenMacDaddy:
Noticed that the autoupdate section used HTTP instead of HTTPS.


Calinou (approved):
Thanks!

PR #665, opened 2022-07-28
cnc-ddraw: Add version 4.4.7.0

RavenMacDaddy:
cnc-ddraw can fix compatibility issues with older games, such as black screen, bad performance, crashes, or defective Alt+Tab. It does this by wrapping DirectDraw to renderers like GDI, OpenGL, or Direct3D 9.

> - [x] stable
> - [ ] beta
> - [ ] dev/nightly/canary


Calinou (approved):
Thanks!

PR #673, opened 2022-08-03
polymc: Use portable version instead

Pato05:

> - [x] stable
> - [ ] beta
> - [ ] dev/nightly/canary


bucket/polymc.json
```diff
@@ -3,19 +3,36 @@
...
+    ],
     "checkver": {
         "github": "https://github.com/PolyMC/PolyMC"
     },
     "autoupdate": {
-        "url": "https://github.com/PolyMC/PolyMC/releases/download/$version/PolyMC-Windows-$version.zip"
+        "url": "https://github.com/PolyMC/PolyMC/releases/download/$version/PolyMC-Windows-Portable-$version.zip"
     }
-}
+}
```
Calinou, line 38:
```suggestion
}

```

Calinou (approved):

dada513:
why?

DioEgizio:
just why

Calinou:
> just why

Scoop aims to use portable apps in manifests whenever possible, as they don't modify files outside the location they're explicitly written in. The issue is that switching from non-portable to portable breaks existing setups, and I'm not sure if it's worth the trouble for scoop-games (where most manifests are non-portable anyway).

Pato05:
> > just why
> 
> Scoop aims to use portable apps in manifests whenever possible, as they don't modify files outside the location they're explicitly written in.
> 
> The issue is that switching a manifest from non-portable to portable breaks existing setups, and I'm not sure if it's worth the trouble for scoop-games (where most manifests are non-portable anyway).

Yeah, I actually didn't think of that before, wouldn't it be possible to migrate from the non-portable build with a pre-install script, maybe?

PR #674, opened 2022-08-04
Revert "polymc: Use portable version instead (#673)"

DioEgizio:
This reverts commit caae8a7bd9b54bb4ae39da696e26b0a9a0f6de66.

Useless, doesn't make any sense and people that used that package before that change won't see their instances anymore with that change in updates

> - [x] stable
> - [ ] beta
> - [ ] dev/nightly/canary


sitiom:
@DioEgizio I'm actually against this change. Scoop uses portable versions if possible. If you don't like that, I suggest you use Winget instead.

DioEgizio:
if you want it using the portable build instead, make a pr that has a pre launch script copying the files, but without this change users would lose access to their instances when updating

PR #677, opened 2022-08-06
polymc: Use portable configuration

sitiom:
Reverts #674.

By principle, Scoop should use portable versions if possible. Otherwise, they should be in the [Nonportable](https://github.com/ScoopInstaller/Nonportable) bucket. This has already been applied to the other official buckets (https://github.com/ScoopInstaller/Extras/pull/7573), the games bucket should follow too.

- [x] I have read the [Contributing Guide](https://github.com/ScoopInstaller/.github/blob/main/.github/CONTRIBUTING.md).

### DIFF
--- a/bucket/polymc.json
+++ b/bucket/polymc.json
@@ -30,8 +30,7 @@
         "themes",
         "translations",
         "accounts.json",
-        "polymc.cfg",
-        "notifications.json"
+        "polymc.cfg"
     ],
     "checkver": {
         "github": "https://github.com/PolyMC/PolyMC"
--- a/bucket/polymc.json
+++ b/bucket/polymc.json
@@ -3,8 +3,13 @@
     "description": "An Open Source Minecraft launcher with the ability to manage multiple instances, accounts and mods. Focused on user freedom and free redistributability.",
     "homepage": "https://polymc.org/",
     "license": "GPL-3.0-only",
-    "url": "https://github.com/PolyMC/PolyMC/releases/download/1.4.1/PolyMC-Windows-1.4.1.zip",
-    "hash": "702d833313318fe8e3dbafc9540d0db23b1a4c6f682283aaa6b302886c5f1aee",
+    "url": "https://github.com/PolyMC/PolyMC/releases/download/1.4.1/PolyMC-Windows-Portable-1.4.1.zip",
+    "hash": "9f363ee10a880364c0010a9c996094ca304eb390c9b966a388990772fbca59ef",
+    "pre_install": [
+        "if (!(Test-Path \"$persist_dir\\accounts.json\")) { New-Item \"$dir\\accounts.json\" | Out-Null }",
+        "if (!(Test-Path \"$persist_dir\\polymc.cfg\")) { New-Item \"$dir\\polymc.cfg\" -Value \"Analytics=false`r`nAutoUpdate=false`r`nIconTheme=pe_colored\" | Out-Null }",
+        "if (!(Test-Path \"$persist_dir\\notifications.json\")) { New-Item \"$dir\\notifications.json\" | Out-Null }"
+    ],
     "bin": "PolyMC.exe",
     "shortcuts": [
         [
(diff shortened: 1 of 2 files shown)


sitiom:
A migration script would keep current users happy (https://github.com/Calinou/scoop-games/pull/674#issuecomment-1207192740).

sitiom:
@Calinou All good. Run the migration script to migrate your config to Scoop:
image

Ping @dada513 @DioEgizio @Pato05 @txtsd

DioEgizio:
polymc-qt5 should be adapted to this too

bucket/polymc.json
```diff
@@ -3,19 +3,40 @@
...
+        "accounts",
+        "assets",
+        "instances",
+        "libraries",
+        "meta",
+        "themes",
+        "translations",
+        "accounts.json",
+        "polymc.cfg",
+        "notifications.json"
```
DioEgizio, line 34: notifications don't exist in polymc anymore for a long time

scripts/polymc/migrate.ps1
```diff
@@ -0,0 +1,28 @@
...
+    "accounts",
+    "assets",
+    "instances",
+    "libraries",
+    "meta",
+    "themes",
+    "translations",
+    "accounts.json",
+    "polymc.cfg",
+    "notifications.json"
```
DioEgizio, line 11: .

sitiom:
> polymc-qt5 should be adapted to this too

@DioEgizio I'm assuming the persist files are the same in the legacy version, right? They seem to be missing in the manifest.

scripts/polymc/migrate.ps1
```diff
@@ -0,0 +1,27 @@
+$files = @(
+    "accounts",
+    "assets",
+    "instances",
+    "libraries",
+    "meta",
+    "themes",
+    "translations",
+    "accounts.json",
+    "polymc.cfg",
```
DioEgizio, line 10:
```suggestion
    "polymc.cfg"
```

I assume

DioEgizio:
> > polymc-qt5 should be adapted to this too
> 
> @DioEgizio I'm assuming the persist files are the same in the legacy version, right? They seem to be missing in that manifest.

Right now polymc-qt5 uses the not portable build, but the files stay in %APPDATA%/PolyMC too, so the migration script should be the same

scripts/polymc/migrate.ps1
sitiom: Oops

sitiom:
> Right now polymc-qt5 uses the not portable build, but the files stay in %APPDATA%/PolyMC too, so the migration script should be the same

Are you sure? It has been using the [portable version](https://github.com/Calinou/scoop-games/blob/3fe019f9b03041fabee8e7b004f6a48ff6578590/bucket/polymc-qt5.json#L6) from the start, so I don't think a migration script is needed for that. I'm just concerned about the lack of persisted files in that manifest.

sitiom:
After some thought, I would prefer using the non-portable config version because if PolyMC is installed globally, you would need admin permissions to write to files every time.
image

Pato05:
@sitiom maybe keeping it non-portable for global installs?

Shouldn't be hard since I saw that PolyMC enables portable mode when there's a "portable.txt" file in its root directory.

It would be better keeping all data in a predictable directory to make uninstallation easier, instead of everything scrambled in the file system.

DioEgizio:
Yes, the actual difference between portable and not portable configurations is managed by the portable.txt file; if it's there is portable, if it's not there it's not portable

sitiom:
> @sitiom maybe keeping it non-portable for global installs?

> It would be better keeping all data in a predictable directory to make uninstallation easier, instead of everything scrambled in the file system.

If you make the global installation non-portable and the local portable, that would be scrambled, wouldn't it?

Pato05:
> > @sitiom maybe keeping it non-portable for global installs?
> 
> > It would be better keeping all data in a predictable directory to make uninstallation easier, instead of everything scrambled in the file system.
> 
> If you make the global installation non-portable and the local portable, that would be scrambled, wouldn't it?

yes but most users would install locally anyways, and you could add a note for that, to warn the user

sitiom:
> this would resolve the problem for local installs

It's not worth the inconsistencies that it will bring between the local and global installations. The current setup works well, config files are located at `%appdata%\PolyMC` regardless of whether it is local or global.

Calinou:
I think it's better to keep polymc non-portable in this case.

PR #680, opened 2022-08-18
gzdoom: Make Save and Screenshots persistent

farsil:
This PR makes the Save and Screenshots persistent, to allow the user to keep their savegames and screenshots between installs. 

> - [x] stable
> - [ ] beta
> - [ ] dev/nightly/canary

### DIFF
--- a/bucket/gzdoom.json
+++ b/bucket/gzdoom.json
@@ -6,7 +6,7 @@
     "notes": [
         "Place WAD files (game data) in:",
         "",
-        "    $persist_dir\\WAD"
+        "    $persist_dir\\..\\_doom"
     ],
     "architecture": {
         "64bit": {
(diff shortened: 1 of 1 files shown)


Calinou:
> It also relocates the WAD directory from persist/_doom to persist/gzdoom/WAD.

This is probably worth discussing in a separate PR, as I intentionally made all Doom source ports use a single `_doom` directory. This allows using the same WAD folder for all source ports (including IWADs, so you don't have to keep several copies of them).

farsil:
> > It also relocates the WAD directory from persist/_doom to persist/gzdoom/WAD.
> 
> This is probably worth splitting to a separate PR, as I intentionally made all Doom source ports use a single `_doom` directory. This allows using the same WAD folder for all source ports (including IWADs, so you don't have to keep several copies of them).

Makes sense, I wasn't aware of multiple source ports available in this repository. I'll remove it from this PR!

Calinou (approved):

PR #700, opened 2022-09-18
Cemu: Update and change release source

BlackPowerade:
The promised day hath come, cemu is now open source.     
Leaving this a draft for now, until the release structure straightens out. A stable and dev version are list on the main page, but the github releases show 2.0 as latest and 2.0-2 as latest dev.

### DIFF
--- a/bucket/cemu-dev.json
+++ b/bucket/cemu-dev.json
@@ -1,11 +1,10 @@
 {
     "version": "2.0-2",
-    "description": "Experimental software to emulate Wii U applications on PC",
+    "description": "A Nintendo WiiU emulator capable of online play (development version)",
     "homepage": "https://cemu.info/",
     "license": "MPL-2.0",
     "suggest": {
-        "Microsoft Visual C++ Runtime 2022": "extras/vcredist2022",
-        "cemuhook": "cemuhook"
+        "Microsoft Visual C++ Runtime 2022": "extras/vcredist2022"
     },
     "architecture": {
         "64bit": {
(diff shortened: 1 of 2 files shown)


RavenMacDaddy:
Holy nuggets, I just missed this completely, spending literally hours over at #701. Haha, damn it…

RavenMacDaddy:
Consider the following:

As per the information by the [Cemuhook author](https://cemuhook.sshnuke.net/), the tool is neither compatible anymore, nor should it be used because Cemu has gone open-source (the same features could be implemented into the project instead).

The corresponding code in the manifest therefore ought to be removed from `suggest` and `installer > script`.

* Add an alias for `bin`.
* Since the architecture property [should only be used when there are both 32- and 64-bit versions available](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties), the manifest ought to only use `url` (the app only compiles for **64-bit**).

BlackPowerade:
Noted. bins updated and cemuhook removed.     
The 64bit prop should stay. It's being used here stop anyone stuck in the dark ages from installing a program that isn't compatible. When the `32|64bit` prop is missing, scoop installs regardless of architecture. When it isn't, scoop will only use the props your system supports. This works both ways. Just like how `64bit` only won't install on a 32bit, a manifest with only `32bit` won't install on 64bit.

RavenMacDaddy:
Good points - and great work.

BlackPowerade:
No, bin will alias as-is without saying what you want it to be.    
just `"bin": "cemu.exe"` means you have to type cemu.exe to invoke from the command line.    
It's a little redundant but its nice to call a program without a file extension. Also more in line with linux and macos.

RavenMacDaddy:
I'm probably misunderstanding, but I don't see an alias committed, so it's literally saying `"bin": "cemu.exe"` right now as mentioned – so that means it's currently obligatory to type out `cemu.exe`, right?

BlackPowerade:
Yeah, you're right. I forgot to commit and push them.

Calinou:
Is this good to merge now? Just asking to make sure :slightly_smiling_face:

BlackPowerade:
@Calinou Go ahead and merge.    
We'll have to keep a close eye on the manifests. Cemu will undoubtably go through a lot of development now that anyone can work on it. Time will tell what changes it will bring.

PR #701, opened 2022-09-18
cemu-dev: Removes deprecated features - updates every property to what's current

RavenMacDaddy:
As per the information by the [Cemuhook author](https://cemuhook.sshnuke.net/), the tool is neither compatible anymore, nor should it be used since Cemu has gone open-source.

Also, when installing 'cemuhook', the tool will rather look for – and install – the _old_ version of Cemu if it isn't installed.

The corresponding code is therefore removed from `suggest` and `installer > script`.

---

Other changes:

* As per the [license](https://github.com/cemu-project/Cemu/blob/main/LICENSE.txt) on the emulator's repository, the `license` property is updated to reflect its usage of the _Mozilla Public License Version 2.0_.
* Since the `architecture` property [should only be used when there are both 32- and 64-bit versions available](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifests#optional-properties), the manifest is now using `url` only, as intended.
* Adds alias for `bin`.
* Updates the `checkver` URL to the one used by the authors of Cemu - along with the corresponding `regex` string.
* `autoupdate` is updated to reflect these changes.

---

### DIFF
--- a/bucket/cemu-dev.json
+++ b/bucket/cemu-dev.json
@@ -1,30 +1,20 @@
 {
-    "version": "2.0",
+    "version": "2.0-2",
     "description": "Nintendo Wii U emulator",
     "homepage": "https://cemu.info/",
     "license": {
-        "identifier": "Freeware",
-        "url": "https://cemu.info"
+        "identifier": "MPL-2.0",
+        "url": "https://github.com/cemu-project/Cemu/blob/main/LICENSE.txt"
     },
     "suggest": {
-        "cemuhook": "cemuhook",
         "Microsoft Visual C++ Runtime 2022": "extras/vcredist2022"
     },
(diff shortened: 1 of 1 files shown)


RavenMacDaddy:
Haha, that reference was obviously not intended.

RavenMacDaddy:
Gosh darn it… Hmm, how do I fix that…

Going to take a walk and think.

RavenMacDaddy:
Well…out of ideas for the moment.

RavenMacDaddy:
Sure was…

Can both the `$version` and `$preReleaseVersion` variables be used at the same time somehow to use the latest type that's available?

RavenMacDaddy:
Closing this to check #700 instead.

PR #716, opened 2022-10-12
romsorter: Update URL/Autoupdate

RavenMacDaddy:
Also updates grammar and the like in the `description` and `notes`.

The latest download URL of the app uses its version at the tail end.

I've also created [an issue](https://github.com/drakewill-CRL/ROMSorter/issues/5) regarding portable mode, as the config of the app is currently created in **Local AppData**.


Calinou (approved):
Thanks!

PR #734, opened 2022-11-01
prismlauncher: Declare architecture

DioEgizio:
since 5.1, Windows is 64 bit and Windows-Legacy is 32 bit


Calinou:
64-bit-only should be declared as such, but 32-bit-only should not as it'll cause installation to fail on 64-bit systems (last time I checked).

DioEgizio:
done

PR #749, opened 2022-11-19
Switch for legacy

onsamyj:
The `latest` is actually a development build, that need a different manifest (this one doesn’t work: shim, persist, etc.).

Creating an additional manifest in `scoop-games` for `duckstation-preview` (or something like that) can be an option, if someone wants to do that.

OR

This manifest need to be reworked and `duckstation-legacy` created as an option.

Closes #XXXX

Relates to #XXXX

### DIFF
--- a/bucket/duckstation.json
+++ b/bucket/duckstation.json
@@ -11,8 +11,8 @@
         "Place the BIOS file in $persist_dir\\bios",
         "Learn more at: https://www.duckstation.org/wiki/BIOS"
     ],
-    "url": "https://github.com/stenzek/duckstation/releases/download/latest/duckstation-windows-x64-release.zip",
-    "hash": "e52140057eea1c4a4218a22a3ba52af1051cedc468a4eda9d218ccdc7bebddf4",
+    "url": "https://github.com/stenzek/duckstation/releases/download/legacy/duckstation-windows-x64-release.zip",
+    "hash": "486bab13e3bb6ac19bd9145558afda0486613898d41e559feb8c8963a76229b4",
     "installer": {
         "script": [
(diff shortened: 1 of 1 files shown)


Calinou:
I think we can add a `duckstation-dev` that reflects the latest development build, and switch this manifest over to the legacy build.

bucket/duckstation.json
```diff
@@ -11,8 +11,8 @@
         "Place the BIOS file in $persist_dir\\bios",
         "Learn more at: https://www.duckstation.org/wiki/BIOS"
```
Calinou, line 12: This URL is 404 – is there a suitable replacement for it?

Calinou (approved):
Thanks!

Calinou:
On second thought, I wonder if the development build is actually intended to be the "main" build for most users. Many emulators work like that.

@stenzek Is the development build intended for general usage yet? I see its release tag indicates it as a stable release, rather than as a pre-release.

onsamyj:
Until we know, switch to `legacy` means that nothing going to be broken. There are plenty of emulators that have _stable_, _alpha_, _nightly_, whatever else versions, so I don’t see this as a problem – we just need a manifest for another version.

There is a blurb [on the main page](https://github.com/stenzek/duckstation/) about BIOS:

> A "BIOS" ROM image is required to to start the emulator and to play games. You can use an image from any hardware version or region, although mismatching game regions and BIOS regions may have compatibility issues. A ROM image is not provided with the emulator for legal reasons, you should dump this from your own console using Caetla or other means.

Is that enough?

Calinou:
> Is that enough?

Yes, I suppose that paragraph could be copy-pasted into the `notes` section.

Issue #751, opened 2022-11-23
duckstation: Can not create shim


ppvnf:
I'm getting this error when trying to install it:

![Captura de tela 2022-10-25 231259](https://user-images.githubusercontent.com/73800734/197918426-7a2f5712-1bf4-4ff3-9647-2e7101a2a0d6.png)

PR #752, opened 2022-11-23
duckstation: Fix installation

ppvnf:
Closes https://github.com/Calinou/scoop-games/issues/751

### DIFF
--- a/bucket/duckstation.json
+++ b/bucket/duckstation.json
@@ -21,13 +21,6 @@
             "}"
         ]
     },
-    "bin": [
-        [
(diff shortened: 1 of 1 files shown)


Calinou:
Is there no more binary we could add to the `PATH`? I like having programs available in the `PATH` for easier access, especially if they have a command line interface that is worth using for automation purposes.

ppvnf:
I guess the regular binary "duckstation-qt-x64-ReleaseLTCG.exe" supports CL parameters

Calinou (approved):
Anyway, let's merge this for now to fix installation. We can look into readding a binary path later.

PR #757, opened 2022-12-02
vkquake: Remove `extract_dir` to fix installation

RavenMacDaddy:
The latest version deprecates the usage of an `extract dir` (based on the respective contents of the archives for each architecture).

Removing these from the manifest results in a successful install – this change reflects that.


Calinou (approved):
Thanks!

PR #764, opened 2022-12-18
pegasus: Fix `checkver` to be more reliable for version checking

Zliced13:
Reason: In my testing on my computer: The previous checkver in this manifest did not always get the right version, or returned a regex error. This new checkver is always consistent, and it returns the correct version every time,


Calinou (approved):
Thanks!

PR #778, opened 2023-01-04
dgvoodoo2: Add shorcut

onsamyj:
You can use one Control Panel to configure all the games you are using dgVoodoo2 with. So having a shortcut to it make sense.

Closes #XXXX

Relates to #XXXX


Calinou (approved):
Thanks!

PR #786, opened 2023-01-13
pcsx2-dev: Remove unnecessary persist entry

mnxn:
It looks like PCSX2 has stopped using the `shaders\GS_FX_Settings.ini` file to store settings, causing an empty file to be created there when scoop installs the program. I believe the configuration has been moved to `inis\GS.ini` which is already persisted due to the entry for the `inis` directory.

Removing this line stops the unnecessary file from being created.


Calinou (approved):
Thanks!

PR #829, opened 2023-02-26
raine: Update to version 0.95

RavenMacDaddy:
Also reorders Autoupdate section to have the same order as the installation section.

### DIFF
--- a/bucket/raine.json
+++ b/bucket/raine.json
@@ -1,17 +1,17 @@
 {
-    "version": "0.94.12b",
+    "version": "0.95",
     "description": "RAINE is an emulator for arcade games",
     "homepage": "https://raine.1emulation.com/",
     "license": "Artistic-2.0",
     "architecture": {
         "32bit": {
             "url": [
-                "https://raine.1emulation.com/archive/raine32-0.94.12b.7z",
-                "https://raine.1emulation.com/archive/dlls32-0.92x.7z"
+                "https://raine.1emulation.com/archive/raine32-0.95.7z",
+                "https://raine.1emulation.com/archive/dlls32-0.95.7z"
             ],
             "hash": [
-                "58c7727dce216c9ef1476a37a739a24e4ee3d5248517e14fe56bdb759eda9469",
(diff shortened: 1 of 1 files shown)


RavenMacDaddy:
Does anyone see the culprit?

RavenMacDaddy:
I really don't know any more...no errors in IDE, formatted with tools in NP++, Autoupdate is now LITERALLY the same as before the commit, but it doesn't work.

Insight is greatly appreciated.

RavenMacDaddy:
…and now changes become empty? What's actually going on, did the hub just break on itself? I'll gladly be wrong about everything, I just don't see it.

bucket/raine.json
```diff
@@ -1,80 +1,80 @@
...
+		"demos",
+		"emudx",
+		"roms",
+		"savedata",
+		"savegame",
+		"screens"
+	],
+	"checkver": {
+		"url": "https://raine.1emulation.com/download/latest.html",
+		"regex": "Available files for version ([\\d.]+[a-zA-Z]?)"
```
kazzarin, line 68: checkver is reading `0.95.0` from the site, while the binaries use `0.95`
```suggestion
		"regex": "/archive/raine32-([\\d.]+[a-zA-Z]?)\\.7z"
```

kazzarin (commented):
The current tests are failing because it prefers spaces over tabs.

RavenMacDaddy:
> The current tests are failing because it prefers spaces over tabs.

Thank you, I'll get right to it. Also, appreciate the regex fix @kazzarin.

RavenMacDaddy:
Did I read right that the test is actually failing on the `racelab` manifest, rather than this commit?

Calinou:
> Did I read right that the test is actually failing on the `racelab` manifest, rather than this commit?

Yes; this should be fixed after rebasing this PR now that I merged https://github.com/Calinou/scoop-games/pull/833.

Calinou (approved):
Thanks!

PR #831, opened 2023-02-27
flycast: Update to version 2.1

makamys:
Version 2.1 uses a capitalized V in the version tag, unlike previous releases. This caused autoupdate to fail.

I changed the autoupdate url to use a capital V, ran checkver, then reverted it to an uncapitalized v, since I assume this was a one-time mistake, and future releases will use the uncapitalized letter again.

### DIFF
--- a/bucket/flycast.json
+++ b/bucket/flycast.json
@@ -1,5 +1,5 @@
 {
-    "version": "2.0",
+    "version": "2.1",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

makamys:
By the way, the error in the test is unrelated, and is caused by the leading space Excavator added to line 9 of racelab.json an hour ago (https://github.com/Calinou/scoop-games/commit/8052a569ee3ebb38b369d57a742b83fc1d213434)

PR #873, opened 2023-04-18
uvtt2fgu: Remove unused dependencies

LoZeno:
After talking with the original developer of the tool, it looks like the executable does _not_ need python to run (it's generated with pyinstaller). The confusion was born out of the fact that the instructions in the [README ](https://github.com/Imagix/uvtt2fgu) mention:
`uvtt2fgu.py requires a python3 installation with PIP.`
But that is true only if running the python script from the source code. The executable has all the dependencies embedded.


Calinou (approved):
Thanks!

PR #879, opened 2023-04-28
(chore): Ensure vscode saves newlines

kazzarin:
Without this option, vscode removes newlines which causes tests to fail.


Calinou (approved):
Thanks!

PR #883, opened 2023-04-30
gog-galaxy: Add version 2.0.62.26

brian6932:


brian6932:
>Generic utilities related to gaming (such as Steam or GOG-related utilities) should be added to [Extras](https://github.com/ScoopInstaller/Extras) instead, as they tend to appeal to a wider audience compared to this bucket.

My bad

PR #981, opened 2023-06-03
prismlauncher/prismlauncher-qt5@7.1: start using portable binaries

getchoo:
this implements the portable versions of prismlauncher packages into scoop, which allows for the use of `persist` directories, as well as having both versions installed at the same time with separate data directories. for backwards compatibility, `%APPDATA%\PrismLauncher` is also copied to `persistdir` during a fresh install/when no files are already in `persistdir`

### DIFF
--- a/bucket/prismlauncher-qt5.json
+++ b/bucket/prismlauncher-qt5.json
@@ -3,12 +3,37 @@
     "description": "An open source Minecraft launcher with the ability to manage multiple instances, accounts and mods. Focused on user freedom and free redistributability. Qt 5 build that still supports legacy Windows like 8.1 and 7.",
     "homepage": "https://prismlauncher.org/",
     "license": "GPL-3.0-only",
-    "notes": "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
-    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/7.1/PrismLauncher-Windows-MSVC-Legacy-7.1.zip",
-    "hash": "904efed4e15d0274f25f6fd2bdb9cb61dbbef2ad90dfc9761abb3612869df89f",
+    "notes": [
+        "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
+        "",
+        "This package is now using the portable version of Prism Launcher, and data should have been migrated automatically.",
+        "If you are using a global install on a system with more than one user, you will need to copy a user's data from %appdata% to the new Scoop PrismLauncher persist directory"
+    ],
+    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/7.1/PrismLauncher-Windows-MSVC-Legacy-Portable-7.1.zip",
+    "hash": "4c8217ad3f37c5fe1df52cc7b1018f73a4cb59aea4a4b7c7924f3398c1de2745",
     "suggest": {
         "Microsoft Visual C++ Redistributables": "extras/vcredist2022"
     },
+    "pre_install": [
+        "$migration = $true",
+        "$visibleUserAccounts = Get-LocalUser | Where-Object { $_.Enabled -eq $true }",
+        "if ($global -and $visibleUserAccounts.Count -gt 1) { $migration = $false }",
+        "",
+        "$appdataPath = \"$HOME\\AppData\\Roaming\\PrismLauncher\"",
+        "if ((Test-Path -Path $appdataPath\\*) -and (!(Test-Path -Path $persist_dir\\*)) -and $migration) {",
+        "    Write-Warning \"Migrating data from $appdataPath to $persist_dir (this may take a while)\"",
+        "    New-Item -Type Directory -Path $persist_dir | Out-Null",
+        "    Copy-Item -Recurse -Force $appdataPath\\* $persist_dir\\",
+        "} elseif (!($migration)) {",
+        "    Write-Warning \"A global Scoop installation was detected with multiple user accounts. Please see the notes at the end of the install process.\"",
+        "}",
+        "",
+        "'accounts.json', 'metacache', 'prismlauncher.cfg' | ForEach-Object {",
+        "    if (!(Test-Path -Path $persist_dir\\$_)) {",
+        "        New-Item -Type File $dir/$_ | Out-Null",
+        "    }",
+        "}"
+    ],
     "post_install": [
         "'install-associations', 'uninstall-associations' | ForEach-Object {",
         "    if (Test-Path \"$bucketsdir\\games\\scripts\\prismlauncher-qt5\\$_.reg\") {",
@@ -26,10 +51,25 @@
             "Prism Launcher"
         ]
     ],
+    "persist": [
+        "assets",
+        "cache",
+        "icons",
+        "instances",
+        "libraries",
+        "logs",
+        "meta",
+        "mods",
+        "themes",
+        "translations",
+        "accounts.json",
+        "metacache",
+        "prismlauncher.cfg"
+    ],
     "checkver": {
         "github": "https://github.com/PrismLauncher/PrismLauncher"
     },
     "autoupdate": {
-        "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/$version/PrismLauncher-Windows-MSVC-Legacy-$version.zip"
+        "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/$version/PrismLauncher-Windows-MSVC-Legacy-Portable-$version.zip"
     }
 }
--- a/bucket/prismlauncher.json
+++ b/bucket/prismlauncher.json
@@ -3,16 +3,41 @@
     "description": "An open source Minecraft launcher with the ability to manage multiple instances, accounts and mods. Focused on user freedom and free redistributability.",
     "homepage": "https://prismlauncher.org/",
     "license": "GPL-3.0-only",
-    "notes": "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
+    "notes": [
+        "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
+        "",
+        "This package is now using the portable version of Prism Launcher, and data should have been migrated automatically.",
+        "If you are using a global install on a system with more than one user, you will need to copy a user's data from %appdata% to the new Scoop PrismLauncher persist directory"
+    ],
     "architecture": {
         "64bit": {
-            "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/7.1/PrismLauncher-Windows-MSVC-7.1.zip",
-            "hash": "45bc1298b2bb55f6f44cb8420667085b8c1d417030bfd846f671bc6b50e383bf"
+            "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/7.1/PrismLauncher-Windows-MSVC-Portable-7.1.zip",
+            "hash": "222c720f7bdcef97426a22924908bc60dd8a76f0205ee3f1a98a5a1b6b1afd38"
         }
     },
     "suggest": {
         "Microsoft Visual C++ Redistributables": "extras/vcredist2022"
     },
+    "pre_install": [
+        "$migration = $true",
+        "$visibleUserAccounts = Get-LocalUser | Where-Object { $_.Enabled -eq $true }",
+        "if ($global -and $visibleUserAccounts.Count -gt 1) { $migration = $false }",
+        "",
+        "$appdataPath = \"$HOME\\AppData\\Roaming\\PrismLauncher\"",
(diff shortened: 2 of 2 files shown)


he3als (changes requested):
Only changes I'd say is (for global installs):
1. Instead of automatically migrating for global installs (as there's multiple user accounts and it's up to the user to decide which one they want to migrate), instead:
   1. Check if there's Prism Launcher in the AppData recursively in each user
   2. If so, display a warning saying that they will have to manually migrate to the Scoop folder, and if they still want a global install, to either use a non-portable version (not made yet of course) or use `winget`
   3. Alternatively, have a manual migration script included, which would be interactive for which user to copy from
   4. If you don't want to conditionally say a warning, just add it to bottom or top of regular Scoop notes and skip the whole detection
2. For global installs, manually set permissions (potentially with `icacls` or native PowerShell) so that regular users have full control, as Prism Launcher will otherwise fail to create logs in write-protected directories

getchoo:
> Add a note to the bottom or top of regular Scoop notes, with information about manual migration

done!

> To make manual migration easier, you could include an interactive PowerShell script to be manually ran

frankly i'm not sure if this is a good idea or really needed in the first place. this would be an entire script for something that is honestly an easy transition (copy one folder's content to another) and the current pre-install script can cover a vast majority of cases (installing at a user level and upgrading to this version).

> For global installs, manually set permissions (potentially with icacls or native PowerShell) so that regular users have full control, as Prism Launcher will otherwise fail to create logs in write-protected directories, meaning it won't launch

i'm not exactly sure what permissions to set here. i could make it open to everyone, but i don't think that would be a good idea as it would allow for arbitrary editing of files. i could also set it to be accessible for specific users when they run the exe, but it would be near impossible to tell which user(s) it should be. this kinda goes with my idea that global installs should not be handled by this package, as the prismlauncher portable exe just doesn't fit with it well. i feel a better alternative for these use cases would be another package in the non-portable bucket or using a version of prism from a different package manager such as choco or winget, where the non-portable version is distributed.

> Making a note about the manifest being portable now

done!

he3als:
> i'm not exactly sure what permissions to set here. i could make it open to everyone, but i don't think that would be a good idea as it would allow for arbitrary editing of files. i could also set it to be accessible for specific users when they run the exe, but it would be near impossible to tell which user(s) it should be. this kinda goes with my idea that global installs should not be handled by this package, as the prismlauncher portable exe just doesn't fit with it well. i feel a better alternative for these use cases would be another package in the non-portable bucket or using a version of prism from a different package manager such as choco or winget, where the non-portable version is distributed.

You could block global installs, although I think that a better idea would be to just set it so that the `Users` group has full access. yet it doesn't apply recursively, so that executables can't be overwritten (for security), but log files can be read, created, etc...

bucket/prismlauncher-qt5.json
```diff
@@ -3,12 +3,27 @@
...
+    ],
+    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/6.3/PrismLauncher-Windows-MSVC-Legacy-Portable-6.3.zip",
+    "hash": "521804443e78a68cdc62f6f1645756f8d74a15ab1941c7ff48a7deb37d419ee3",
     "suggest": {
         "Microsoft Visual C++ Redistributables": "extras/vcredist2022"
     },
+    "pre_install": [
+        "$appdataPath = \"$HOME\\AppData\\Roaming\\PrismLauncher\"",
+        "if ( (Test-Path -Path $appdataPath\\*) -and (!(Test-Path -Path $persist_dir\\*)) -and (!($global)) ) {",
+        "    Write-Host \"Migrating data from $appdataPath to \"$persist_dir\" (this may take a while)\"",
```
he3als, line 18: The `\"$persist_dir\"`'s quotes are not interpreted as quotes

vmware_XE0dRuv3uo

Also, it's probably best to make it a warning or coloured, just so that it's easier to see in the Scoop output.

bucket/prismlauncher-qt5.json
```diff
@@ -3,12 +3,27 @@
     "description": "An open source Minecraft launcher with the ability to manage multiple instances, accounts and mods. Focused on user freedom and free redistributability. Qt 5 build that still supports legacy Windows like 8.1 and 7.",
     "homepage": "https://prismlauncher.org/",
     "license": "GPL-3.0-only",
-    "notes": "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
-    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/6.3/PrismLauncher-Windows-MSVC-Legacy-6.3.zip",
-    "hash": "003b54560935c341ab825d41962323e56d649ea15a29f69d4ce64b93326c866c",
+    "notes": [
```
he3als, line 6: Maybe add a note about the migration not deleting the old AppData directory? As well as maybe making the notes more concise:
```
        "Run this for file associations with .ZIPs and .MRPACKS: \"$dir\\install-associations.reg\"",
        "This package is now fully portable, with data being the in the Scoop persist directory, and it should have been migrated automatically. If this is a global Scoop install, you will have to manually copy your data from AppData to persist."
```

bucket/prismlauncher-qt5.json
```diff
@@ -3,12 +3,27 @@
...
+        "This package is now using the portable version of Prism Launcher, and data should have been migrated automatically. If you are using a global install, you will need to copy your own data to the new Scoop persist directory"
+    ],
+    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/6.3/PrismLauncher-Windows-MSVC-Legacy-Portable-6.3.zip",
+    "hash": "521804443e78a68cdc62f6f1645756f8d74a15ab1941c7ff48a7deb37d419ee3",
     "suggest": {
         "Microsoft Visual C++ Redistributables": "extras/vcredist2022"
     },
+    "pre_install": [
+        "$appdataPath = \"$HOME\\AppData\\Roaming\\PrismLauncher\"",
+        "if ( (Test-Path -Path $appdataPath\\*) -and (!(Test-Path -Path $persist_dir\\*)) -and (!($global)) ) {",
```
he3als, line 17: Could make it so that if it's a global install, then it says a warning like:
```
Write-Warning "A global Scoop installation was detected with multiple Windows user accounts, therefore this manifest can't determine which user-specific Prism Launcher data to migrate, so it has to be done manually
```
Alternatively, there could be detection for only one account on Windows, and if so, it migrates.

he3als (changes requested):

he3als:
https://github.com/PrismLauncher/PrismLauncher/issues/136 The logs will now be in their own folder.

Once the next Prism Launcher version is released, this will work without the permissions issues

bucket/prismlauncher-qt5.json
he3als, line 18: The `\"$persist_dir\"`'s quotes are not interpreted as quotes

vmware_XE0dRuv3uo

Also, it's probably best to make it a warning or coloured, just so that it's easier to see in the Scoop output.

bucket/prismlauncher-qt5.json
he3als, line 17: Instead of completely blocking migration for global installs, I think that it's best to use implement either:
- A check to see if there's one user account in Windows, if so, migrate
- Recursively check the folders of `C:\Users` to see if there's more than one `\AppData\Roaming\PrismLauncher` folder, if there's only one found, migrate
  - Permissions could be an issue though, not sure if users can read other user data easily like that by default

However, if it's determined that it can't auto-migrate, then a warning should be displayed:
```powershell
Write-Warning "A global Scoop installation was detected with multiple user accounts. Therefore, it can't be determined which account to automatically migrate Prism Launcher data from, so you will need to manually migrate your Prism Launcher data from AppData."
```

bucket/prismlauncher-qt5.json
```diff
@@ -26,10 +41,24 @@
             "Prism Launcher"
         ]
     ],
+    "persist": [
```
he3als, line 44: Prism Launcher 7.1 now has a `logs` folder, this should be added to `persist`.

bucket/prismlauncher-qt5.json
```diff
@@ -3,12 +3,27 @@
...
     "homepage": "https://prismlauncher.org/",
     "license": "GPL-3.0-only",
-    "notes": "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
-    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/6.3/PrismLauncher-Windows-MSVC-Legacy-6.3.zip",
-    "hash": "003b54560935c341ab825d41962323e56d649ea15a29f69d4ce64b93326c866c",
+    "notes": [
+        "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
+        "This package is now using the portable version of Prism Launcher, and data should have been migrated automatically. If you are using a global install, you will need to copy your own data to the new Scoop persist directory"
+    ],
+    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/6.3/PrismLauncher-Windows-MSVC-Legacy-Portable-6.3.zip",
```
he3als, line 10: Prism Launcher 7.1 released.

bucket/prismlauncher-qt5.json
```diff
@@ -3,12 +3,27 @@
     "description": "An open source Minecraft launcher with the ability to manage multiple instances, accounts and mods. Focused on user freedom and free redistributability. Qt 5 build that still supports legacy Windows like 8.1 and 7.",
     "homepage": "https://prismlauncher.org/",
     "license": "GPL-3.0-only",
-    "notes": "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
-    "url": "https://github.com/PrismLauncher/PrismLauncher/releases/download/6.3/PrismLauncher-Windows-MSVC-Legacy-6.3.zip",
-    "hash": "003b54560935c341ab825d41962323e56d649ea15a29f69d4ce64b93326c866c",
+    "notes": [
+        "To add Prism Launcher file association options for .ZIPs and .MRPACKS, run this: \"$dir\\install-associations.reg\"",
+        "This package is now using the portable version of Prism Launcher, and data should have been migrated automatically. If you are using a global install, you will need to copy your own data to the new Scoop persist directory"
```
he3als, line 8: I think this would be better if it was reworded a bit:
```json
        "This package is now using the portable version of Prism Launcher, and data should have been migrated automatically."
        "If this package can't determine which data to migrate, you will need to manually copy your data from AppData to the new Scoop PrismLauncher persist directory."
```

he3als (changes requested):
Some general changes. 

Prism Launcher 7.1 has released, and it adds a logs directory for logs. This means that you can now persist logs, so there's no permission issues with global installs of Prism Launcher trying to write logs to the `prismlauncher` app directory.

bucket/prismlauncher-qt5.json
getchoo: it wasn't meant to be in quotes, not sure why i did that lol. i am using `Write-Warning` now though

bucket/prismlauncher-qt5.json
getchoo: i don't think either of these are really a good a idea. it isn't very hard to copy things from appdata manually, and both of these add a lot of unneeded complication (to an already somewhat complicated preinstall script)

bucket/prismlauncher-qt5.json
getchoo: done!

bucket/prismlauncher-qt5.json
getchoo: i adopted some of these suggestions as well :)

getchoo:
the logs directory for global installs should work now. i set the permissions to "everyone" as nothing very sensitive to specific users is in them anyways

he3als:
Shouldn't the persist directory be fully writable anyways?

bucket/prismlauncher-qt5.json
```diff
@@ -3,12 +3,43 @@
...
+        "    New-Item -Type Directory -Path $persist_dir | Out-Null",
+        "    Copy-Item -Recurse -Force $appdataPath\\* $persist_dir\\",
+        "}",
+        "",
+        "$logDir = \"$original_dir\\logs\"",
+        "if ((!(Test-Path -Path $logDir))) {",
+        "    New-Item -Type Directory -Path $logDir | Out-Null",
+        "}",
+        "",
+        "$acl = Get-Acl $logDir",
```
he3als, line 29: Setting permissions on the logs folder shouldn't be needed for global installs. To my knowledge, the Scoop `persist` folder is already writable to users, otherwise things like making instances, accounts, etc... wouldn't work as well.

getchoo: massive brain fart...

bucket/prismlauncher-qt5.json
getchoo: this is actually really good, i'll add this now :)

bucket/prismlauncher-qt5.json
he3als: Lol, it's fair enough, I was being confusing about it to be fair.

getchoo:
i think this is all good for a final review now :)

he3als (approved):

PR #1047, opened 2023-10-25
jazz2-resurrection: Update to version 2.1.0

RavenMacDaddy:
Changes to the manifest include:

- Changing the license to GPL 3.0 only as I don't see any mention of GPL 3.0 or later
- Updates the notes to reflect the current setup process, as well as general improvements to the contents
- Adds the architecture property for extract_dir, providing both x32 and x64 versions of the app
- Provides shortcuts to all versions of the app; standard, AVX2 and SDL2
- Persists both the Source dir (where you put all the game data) and the config file (automatically created through pre_install if one doesn't exist), making the installation completely portable

### DIFF
--- a/bucket/jazz2-resurrection.json
+++ b/bucket/jazz2-resurrection.json
@@ -1,48 +1,64 @@
 {
-    "version": "0.7.3",
+    "version": "2.1.0",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #1104, opened 2024-03-08
ps3-system-software: fix autoupdate

Birdie0:
Current regex missed latest [4.91](https://www.playstation.com/en-us/support/hardware/ps3/system-software/) update because of different case in title, added case-insensitive flag to prevent that in future.


Calinou (approved):

PR #1118, opened 2024-03-26
special-k: Add version 23.9.10.1

FlawlessCasual17:
Adds Special K, “the Swiss army knife of PC gaming” to this scoop bucket.

### DIFF
--- a/bucket/special-k-np.json
+++ b/bucket/special-k-np.json
@@ -1,0 +1,56 @@
+{
+    "version": "23.9.10.5",
+    "homepage": "https://www.special-k.info/",
+    "description": "Lovingly referred to as the Swiss Army Knife of PC gaming, Special K does a bit of everything.",
+    "license": "GPL-3.0",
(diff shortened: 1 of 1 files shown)


Calinou:
I wonder if this would be better part of the Extras bucket instead, so that it's in the same location as tools like RTSS.

FlawlessCasual17:
Do you think it should go in the extras bucket?

FlawlessCasual17:
I don't mind putting this in another bucket

FlawlessCasual17:
I'm not sure to be honest

Calinou:
> Do you think it should go in the extras bucket?

Yes, I try to keep this bucket strictly about games and game-related tools, but Special-K is general enough that it can be useful to install as a development tool (for troubleshooting system performance or profiling).

FlawlessCasual17:
Maybe not in the **extras** bucket, it should go in the **non-portable** bucket since _Special K_ is pretty difficult to manually install, so I would rather let the installer do all the hard work.

PR #1134, opened 2024-04-14
aliceincradle: Add version 0.24g

lwd-temp:

### DIFF
--- a/bucket/aliceincradle.json
+++ b/bucket/aliceincradle.json
@@ -1,0 +1,22 @@
+{
+  "version": "0.24g",
+  "description": "A soft Ryona 2D action game in which MP is thoroughly absorbed when the heroine is defeated.",
+  "license": "Freeware",
+  "architecture": {
+    "64bit": {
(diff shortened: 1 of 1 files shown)


lwd-temp:
Note: There's only `checkver`, but no `autoupdate` because its download link is provided directly in HTML, not the API. And Scoop does not support find download link by regex other than `checkver` URL.

bucket/aliceincradle.json
```diff
@@ -0,0 +1,22 @@
+{
+  "version": "0.24g",
+  "description": "A soft Ryona 2D action game in which MP is thoroughly absorbed when the heroine is defeated.",
```
Calinou, line 3:
```suggestion
  "description": "A soft Ryona 2D action game in which MP is thoroughly absorbed when the heroine is defeated",
```

Calinou:
Thanks for opening a pull request!

Unfortunately, given the 18+ nature of the game, I can't add it to scoop-games for legal reasons (we don't ask all users of scoop-games whether they're over 18 when they enable the bucket). To avoid any risks, I'll have to reject this pull request.

lwd-temp:
Understand.

Will ask users if they're over 18 in a preinstall script acceptable?

PR #1148, opened 2024-05-13
OpenTTD: fix architecture and remove baseset installation

WenSimEHRP:
The original manifest file has an error. Win64 and win32 versions are swapped thus it will cause errors on 32-bit computers. Furthermore, the game has arm64 builds provided, so I think it is better to fix the architecture.

This manifest is basically identical with [this manifest](https://github.com/WenSimEHRP/OpenTTD-bucket/blob/master/bucket/openttd.json). I just changed the notes.

Baseset installation (i.e. OpenGFX, OpenMSX, and OpenSFX) are removed because the basesets can be downloaded afterwards, and the game stores them in the user documents folder, in which it is the default content storage folder (all screenshots, mods, basesets, etc.) Plus, the user (e.g. many who have the original TTD graphics, musics and sounds OR they installed the game before) may already have the basesets ready so they don't need extra basesets. Letting the user to download OpenGFX, OpenMSX and OpenSFX for an extra time does not make sense to me.

### DIFF
--- a/bucket/openttd.json
+++ b/bucket/openttd.json
@@ -3,43 +3,28 @@
     "description": "Simulation game based upon Transport Tycoon Deluxe",
     "homepage": "https://www.openttd.org/",
     "license": "GPL-2.0-or-later",
+    "notes": [
+        "This install does not come with graphics, sounds or music.",
+        "They could be downloaded from the OpenTTD content delivery network",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #1154, opened 2024-06-02
gzdoom: Update to version 4.12.2 and fix URL

RavenMacDaddy:
- Updates to version 4.12.2
- Omits the "-64bit" part of the URL, since the latest versions only support 64 bit and therefore don't type it out in the filename.

(The last part should be the reason the manifest wasn't updating properly.)


Calinou (approved):
Thanks!

PR #1155, opened 2024-06-08
steam-achievement-manager-syntaxtm: autoupdate fix for 0.6.0, name change

SandeMC:
The zip filename changed, so the automatic updater didn't update. Also changed how it shows up on the start menu


SandeMC:
@Calinou could the original steam achievement manager be considered for deprecation in favor of this one? This is practically just modern SAM that shouldn't have issues with modern windows in the long run

Calinou:
> could the original steam achievement manager be considered for deprecation in favor of this one?

That makes sense. Feel free to open a separate PR for this.

PR #1172, opened 2024-06-29
lim3ds@2115: Fix extract_dir and add it to autoupdate

RavenMacDaddy:
image

This should fix the exception. I also checked that the app still uses this naming scheme in the latest version.


Calinou (approved):
Thanks!

PR #1180, opened 2024-07-04
daggerfallunity: persist mods and gamefiles

mundifex:
Daggerfall Unity expects `.dfmod` files to be stored within a subfolder of StreamingAssets. Failing to persist this folder means users will lose any installed mods any time the app updates.

Also persisting GameFiles in case users want to place Daggerfall's arena2 folder there and create a truly portable install.


mundifex:
Closes #1181

PR #1206, opened 2024-08-18
bedrock-launcher: Add version 2024.8.18.21

sitiom:
An unofficial Minecraft for Windows launcher that enables similar features from the Minecraft Java Edition Launcher.

### DIFF
--- a/bucket/bedrock-launcher.json
+++ b/bucket/bedrock-launcher.json
@@ -1,0 +1,37 @@
+{
+    "version": "2024.8.18.21",
+    "description": "An unofficial Minecraft Bedrock for Windows Launcher",
+    "homepage": "https://bedrocklauncher.github.io/",
+    "license": "GPL-3.0-only",
+    "notes": "See \"Important Notes\" section at https://github.com/BedrockLauncher/BedrockLauncher/releases/latest",
+    "suggest": {
+        "VCRedist": "extras/vcredist2022",
+        ".NET 6.0 Desktop Runtime": "versions/windowsdesktop-runtime-6.0"
+    },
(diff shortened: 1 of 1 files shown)


bucket/bedrock-launcher.json
```diff
@@ -0,0 +1,37 @@
+{
+    "version": "2024.8.18.21",
+    "description": "An unofficial Minecraft Bedrock for Windows Launcher",
+    "homepage": "https://bedrocklauncher.github.io/",
+    "license": "GPL-3.0-only",
+    "notes": "See \"Important Notes\" section at https://github.com/BedrockLauncher/BedrockLauncher/releases/latest",
```
Calinou, line 6: I suggest mentioning this requirement directly in the notes considering how important it is to get the app working:

> Run the launcher ON AN ADMINISTRATOR ACCOUNT. Running it on a normal account and providing Admin rights at the prompt will result in an error and crash of the launcher. Please also ensure you are using a MICROSOFT account on your pc, or your microsoft store is logged in with a microsoft account which has access to the game (Either bought or game pass).

I'm surprised this is even required in the first place though… I've never heard of an app that requires using an administrator account, not just accepting an UAC prompt. It's been over a decade since I logged into an actual administrator account on Windows.

Rayth: Honestly I'm not even sure but we have seen where when people run the launcher with just UAC prompt on a normal account, they get "access denied" error to the minecraft save location.

sitiom:
Bump

Calinou (approved):
Thanks!

Issue #960, opened 2023-05-16
[Request]: Vassal engine


superusercode:
### Prerequisites

- [X] I have searched all issues/PRs to ensure it has not already been reported or fixed.

### Criteria

- [X] Reasonably well-known and widely used
- [X] A game or a game-related tool
- [X] English interface (or at least English documentation)
- [X] Latest stable or development version
- [X] Full version (i.e. not a trial version)
- [X] Fairly standard install (e.g. uses a version-specific download URL, no elaborate pre/post install scripts)

### Name

Vassal Engine

### Description

The Vassal Engine is a game engine for building and playing online adaptations of board games, tabletop games and card games. It allows users to play in real time over a live Internet connection, and also by email. It runs on all platforms, and is free, open-source software. 

### Homepage

https://vassalengine.org/

### Download Link(s)

https://github.com/vassalengine/vassal/releases/download/3.6.17/VASSAL-3.6.17-other.zip

### Some Indication of Popularity/Repute

https://github.com/vassalengine/vassal 291 stars on github
https://en.wikipedia.org/wiki/Vassal_Engine meets wikipedia notability standards

PR #1250, opened 2024-10-19
VASSAL: Add VASSAL engine v3.7.14

finsterdexter:
Adding manifest for VASSAL. This relates to #960

### DIFF
--- a/bucket/vassal.json
+++ b/bucket/vassal.json
@@ -1,0 +1,36 @@
+{
+    "version": "3.7.14",
+    "description": "VASSAL is a game engine for building and playing online adaptations of board games and card games. Play live on the Internet or by email. VASSAL runs on all platforms, and is free, open-source software.",
+    "homepage": "https://vassalengine.org",
+    "license": {
+        "identifier": "LGPL-2.1-only",
+        "url": "https://github.com/vassalengine/vassal/raw/master/LICENSE"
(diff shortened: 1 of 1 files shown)


finsterdexter:
Don't merge this yet, I need to add a pre_install

finsterdexter:
I added installer and uninstaller properties in the manifest. When I test this it does pop up the UAC prompt since even running silently and setting destination to scoop $dir doesn't cause the installer to not try and elevate to admin. Is that an issue, as I think the goal with scoop manifests is to avoid UAC, but I'm not sure of the best way to avoid that here.

Calinou:
> Is that an issue, as I think the goal with scoop manifests is to avoid UAC, but I'm not sure of the best way to avoid that here.

This is a NSIS-based installer according to `file`:

```
$ file VASSAL-3.7.14-windows-x86_64.exe
VASSAL-3.7.14-windows-x86_64.exe: PE32 executable (GUI) Intel 80386 (stripped to external PDB), for MS Windows, Nullsoft Installer self-extracting archive, 7 sections
```

You should be able to remove the (un)installer properties and add `#/dl.7z` at the end of the download URL, similar to [`sauerbraten.json`](https://github.com/Calinou/scoop-games/blob/master/bucket/sauerbraten.json).

PR #1255, opened 2024-10-26
supermodel: add version 0.3a-git-f9cb795

### LINKED ISSUE #681: Discussion: How do we build a manifest for the SEGA Model 3 emulator Supermodel (uses random string for builds)?
Here's an example of a download URL: https://supermodel3.com/Files/Git_Snapshots/Supermodel_0.3a-git-a60f998_Win64.zip

So the `a60f998` portion is what changes with each release to another random string.

The [repository itself](https://supermodel3.com/Files/Git_Snapshots/) has nice dated releases, that part should be easy to figure out - the random string is the issue.

I wonder if what's used in [this manifest](https://github.com/ScoopInstaller/Extras/blob/fccdc94140356911d62a6a7dd5b761bec9bc7671/bucket/mobaxterm.json) would work?

GriekseEi:
Added supermodel package version 0.3a-git-f9cb795.

The versioning string for the newest development versions of Supermodel is random, but the Supermodel download page sorts the builds in order of newest, so we only need to capture the first match to get the newest version. 

Supermodel in Scoop has to be run using a wrapper script that temporarily sets the working directory to the executable location, as Supermodel is hardcoded to load configuration files relative to the executable. As far as I could see, there are no Scoop options yet for being able to set the working directory of an entry in the bin field. 

Closes #681 

### DIFF
--- a/bucket/supermodel.json
+++ b/bucket/supermodel.json
@@ -1,0 +1,29 @@
+{
+    "version": "0.3a-git-f9cb795",
+    "description": "Sega Model 3 Arcade Emulator",
+    "homepage": "https://supermodel3.com/index.html",
+    "license": "GPL-3.0-only",
+    "url": "https://supermodel3.com/Files/Git_Snapshots/Supermodel_0.3a-git-f9cb795_Win64.zip",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #1298, opened 2024-12-21
Ryujinx@1.2.76: fix version string in autoupdate url and update

Nachtalb:
The autoupdate url used a fixed version in the name of the artefact which made it useless and broken. Using `$version` now. 
And update to `1.2.78` cuz otherwise it doesn't pass the checks. 


Calinou (approved):
Thanks!

PR #1291, opened 2024-12-14
shadps4-dev: fix autoupdate url

AndreiVernon:
Closes #1290.

It seems that the url setting for the autoupdate rule was wrong, the architecture property was used incorrectly. I tested this change and it seems to fix the issue.

- [X] I have read the [Contributing Guide](https://github.com/Calinou/scoop-games/blob/master/CONTRIBUTING.md).

Issue #1290, opened 2024-12-14
shadps4-dev is outdated


clin1234:
Current prerelease version: https://github.com/shadps4-emu/shadPS4/releases/tag/Pre-release-shadPS4-2024-12-13-715ac8a

wyndon:
Yeah, as already said in https://github.com/Calinou/scoop-games/pull/1286#issuecomment-2542530282, the older releases are always getting deleted in favor of the new ones, so it's fundamentally incompatible with how scoop works. Each commit triggers a rebuild and replaces the old dev release.

AndreiVernon:
That shouldn't affect scoop at all from what I understand. The manifest should automatically update to the newest update using `checkver`. 

For some reason the update didn't work properly though - the version number was updated but not the url. I'm trying to figure out why that is and will issue a fix.

wyndon:
Because the frequency of the updates is way too fast. As said each commit triggers a rebuild, and each build replaces the old one. There's often multiple commits in a single day : https://github.com/shadps4-emu/shadPS4/commits/main/

From what I'm seeing, the scoop autoupdater runs every 4 hours, but that still doesn't inherently fix the issue and we'll very often face a 404 link.

Calinou:
I think it'd be best to ask the developer to use fixed file names in this case, then we can use the `nightly` version identifier.

wyndon:
> * Fixed by [fix: shadps4 checkver #1322](https://github.com/Calinou/scoop-games/pull/1322).

Apologies for the double notification, should have posted here instead. This PR fixes `shadps4` but not `shadps4-dev`, so this issue is still relevant.

PR #1322, opened 2025-02-09
fix: shadps4 checkver

yehorb:
shadps4 releases are tagged with an additional dot, which is included in
the extracted version when using the github checkver.

Using the url checkver with a small regex cleanup we can extract the
correct version that actually fits nicely into the autoupdate url.

Relates to #1290 #1291 


Calinou (approved):
Thanks!

PR #1328, opened 2025-02-25
Fix BZFlag 2.4.28 hash and switch the vcredist

blast007:
The original BZFlag 2.4.28 installer that I uploaded was corrupted, so the hash that was pulled into scoop-games was incorrect.  I uploaded a corrected installer file and I corrected the hash. The manifest also referenced extra/vsredist2017, which doesn't exist anymore, so I replaced that with extra/vcredist2022.


Calinou (approved):
Thanks!

PR #1331, opened 2025-02-27
Patch raze to use hotfix zip.

thescarydoor1:
The Raze zip changed names (1.11.0 to 1.11.0a) but the version in the url path was unchanged (still 1.11.0), so now the link and the autoupdate pattern are broken.  ~I think the new name is not proper, so it's better to make a one time patch instead of changing the pattern, and have autoupdate take care of the next one~ Updated with a better solution that handles this weird url

### DIFF
--- a/bucket/raze.json
+++ b/bucket/raze.json
@@ -24,7 +24,7 @@
     "persist": "data",
     "checkver": {
         "url": "https://api.github.com/repos/ZDoom/Raze/releases/latest",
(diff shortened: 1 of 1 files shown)


thescarydoor1:
Found a different solution

This is a more complex regex, which captures 1.11.0 in the url and 1.11.0b in the file name and stores them in separate variables. So this will work with the current url, and will hopefully work with future ones as well

Calinou (approved):
Thanks!

PR #1346, opened 2025-03-30
ryujinx: Update to new release repo

kid1412621:
Update to Ryujinx [new release repo](https://github.com/Ryubing/Stable-Releases) since it migrated to [gitlab](https://git.ryujinx.app/ryubing/ryujinx).

- [x] I have read the [Contributing Guide](https://github.com/Calinou/scoop-games/blob/master/CONTRIBUTING.md).

*Maintainer edit: This closes [#1337](https://github.com/Calinou/scoop-games/issues/1337).*

PR #1339, opened 2025-03-15
Ryujinx update new URL

osp709:
Update URLs for ryujinx package

Closes #1337

- [X] I have read the [Contributing Guide](https://github.com/Calinou/scoop-games/blob/master/CONTRIBUTING.md).

Issue #1337, opened 2025-03-15
[Bug]: New link for Ryujinx


osp709:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

games/ryujinx

### Expected/Current Behaviour

Ryujinx update failing due to wrong download link

### Steps to Reproduce

```console
> scoop update ryujinx
ryujinx: 1.2.82 -> 1.2.86
WARN  Scoop uses 'aria2c' for multi-connection downloads.
WARN  Should it cause issues, run 'scoop config aria2-enabled false' to disable it.
WARN  To disable this warning, run 'scoop config aria2-warning-enabled false'.
Updating one outdated app:
Updating 'ryujinx' (1.2.82 -> 1.2.86)
Downloading new version
Starting download with aria2 ...
Download: 03/15 12:07:00 [ERROR] CUID#8 - Download aborted. URI=https://github.com/GreemDev/Ryujinx/releases/download/1.Download: Exception: [AbstractCommand.cc:351] errorCode=3 URI=https://github.com/GreemDev/Ryujinx/releases/download/1.2.Download: Download Results:
Download: gid   |stat|avg speed  |path/URI
Download: ======+====+===========+=======================================================
Download: 5598b3|ERR |       0B/s|C:/Users/omshr/scoop/cache/ryujinx#1.2.86#e80790a.zip
Download: Status Legend:
Download: (ERR):error occurred.
Download: aria2 will resume download if the transfer is restarted.
Download: If there are any errors, then see the log file. See '-l' option in help/man page for details.

ERROR Download failed! (Error 3) Resource was not found
ERROR https://github.com/GreemDev/Ryujinx/releases/download/1.2.86/ryujinx-1.2.86-win_x64.zip
    referer=https://github.com/GreemDev/Ryujinx/releases/download/1.2.86/
    dir=C:\Users\omshr\scoop\cache
    out=ryujinx#1.2.86#e80790a.zip

[...]

PR #1348, opened 2025-03-31
ryubing-ryujinx: init at 1.2.86

yehorb:
Ryujinx project have moved to their own GitLab platform for development.

The current Ryujinx package links to incorrect packages, so updates do not work.

There are already PRs to fix the issue, linked below.

My version is differs in checkver, it uses GitLab instance API to get the latest version. And I also add Canary release package.

Closes [#1337](https://github.com/Calinou/scoop-games/issues/1337)
Relates to [#1346](https://github.com/Calinou/scoop-games/pull/1346) [#1339](https://github.com/Calinou/scoop-games/pull/1339)

### DIFF
--- a/bucket/ryujinx-canary.json
+++ b/bucket/ryujinx-canary.json
@@ -36,7 +36,7 @@
     ],
     "persist": "portable",
     "checkver": {
-        "url": "https://git.ryujinx.app/api/v4/projects/ryubing%2Fryujinx/repository/tags?search=^Canary-",
+        "url": "https://git.ryujinx.app/api/v4/projects/ryubing%2Fryujinx/repository/tags?search=%5ECanary-",
         "jsonpath": "$[0].name",
         "regex": "Canary-([\\d.]+)"
     },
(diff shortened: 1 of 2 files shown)


Calinou:
Looks good to me, but I wonder if we really have to rename `ryujinx` to `ryubing-ryujinx`. The official name of this continuation is still Ryujinx on its website, so I think that's what users will try to install instinctively.

Could you rename the manifests to `ryujinx` and `ryujinx-canary`? This should be good to merge afterwards.

yehorb:
@Calinou, I rebased and squashed the changes into two commits.

I am not sure why the checker is upset.

It states that:

```text
String 'https://git.ryujinx.app/api/v4/projects/ryubing%2Fryujinx/repository/tags?search=^Canary-' does not validate against format 'uri'.
```

But it is a valid URL that opens in a browser without any issues.

It may be upset about the `^` character, but that how tag search is done in the [GitLab API](https://docs.gitlab.com/api/tags/).

yehorb:
I did not know that, thank you for pointing it out. GilLab docs were kind enough to point out how to encode a / character, but did not mention anything about ^ character.

Calinou (approved):
Thanks!

PR #1370, opened 2025-04-11
xenia, xenia-canary: Improve pre/post install scripts

Mahir007:
This PR makes some improvements to install scripts and other miscellaneous bits.

xenia:
- Add vcredist dependency to suggest
- Remove creation of persistent portable.txt in pre_install script
- Changed how xenia.config.toml is generated (avoid redundancy if appdata migrated)
- Create portable.txt in $dir instead via post_install script

xenia-canary:
- Add vcredist dependency to suggest
- Remove creation of persistent portable.txt as app is inherently portable
- Renamed xenia.config.toml to xenia-canary.config.toml if appdata is migrated
- Changed how xenia-canary.config.toml is generated (avoid redundancy if appdata migrated)
- Removed persisting of now non-existent portable.txt
- Updated autoupdate to include architecture property
 

### DIFF
--- a/bucket/xenia-canary.json
+++ b/bucket/xenia-canary.json
@@ -6,6 +6,9 @@
         "identifier": "BSD-3-Clause",
         "url": "https://github.com/xenia-canary/xenia-canary/blob/canary_experimental/LICENSE"
     },
+    "suggest": {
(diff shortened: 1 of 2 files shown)


Calinou (approved):
Thanks!

PR #1404, opened 2025-05-18
chiaki-ng: Add arm64 build, switch x64 to MSYS2 build

Mahir007:
This PR does the following:
- Adds an arm64 build
- Switches the x64 build to the MSYS2 version (as VC version no longer exists)
- Fixes licence


Calinou (approved):
Thanks!

PR #1431, opened 2025-06-01
melonds: fix configuration persistence

leaumar:
Closes #1430

- [x] I have read the [Contributing Guide](https://github.com/Calinou/scoop-games/blob/master/CONTRIBUTING.md).

Issue #1430, opened 2025-06-01
[Bug]: melonds configuration is in melonDS.toml now, not melonDS.ini


leaumar:
### Prerequisites

- [x] I have written a descriptive issue title.
- [x] I have searched all issues/PRs to ensure it has not already been reported or fixed.
- [x] I have verified that I am using the latest version of Scoop and corresponding bucket.

### Package Name

melonds

### Expected/Current Behaviour

I have not _encountered_ the bug yet per se, I'm just checking out the package and making sure I won't run into future problems with persistence if I switch from a portable copy of desmume to melonds from scoop.

The melonds package lists only `melonDS.ini` for persistence, but that file is 0KB after setting up my controller and running a game. Instead, there's a `melonDS.toml` file next to it that does have the expected content. It seems obvious to me that should be persisted alongside the .ini, this was probably a recent change by the melonds dev that escaped the package maintainer.

I don't know if anything further needs to happen for this "transition" or if it'll _just work_ by updating the config.

PR incoming.

![Image](https://github.com/user-attachments/assets/72a42dc5-5457-4629-9cea-b52402d9280c)

![Image](https://github.com/user-attachments/assets/9f40125f-7869-4162-b3ea-3458dbe9774a)

### Steps to Reproduce

```console
- install melonds
- run melonds, configure anything
- upgrade -> I assume your config will be lost because it isn't persisted
```

### Possible Solution

add melonDS.toml to `persist`

### Scoop and Buckets Version

```console
Current Scoop version:
859d1db5 (HEAD -> master, tag: v0.5.2, origin/master, origin/HEAD) chore(release): Bump to version 0.5.2 (#6080)

'main' bucket:
2a8e19094 (HEAD -> master, origin/master, origin/HEAD) dynamorio: Update to version 11.90.20236

'games' bucket:
310aa4b33 (HEAD -> master, origin/master, origin/HEAD) pcsx2-dev: Update to version 2.3.387
```

### Scoop Config

```console
lastupdate          SCOOP_REPO                           SCOOP_BRANCH last_update
----------          ----------                           ------------ -----------
11/04/2022 17:23:36 https://github.com/lukesampson/scoop master       31/05/2025 10:18:23
```

### PowerShell Version

```console
Name                           Value
----                           -----
PSVersion                      7.2.24
PSEdition                      Core
GitCommitId                    7.2.24
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
[...]

PR #1439, opened 2025-06-03
melonds: fix incorrect persistence logic

leaumar:
@Calinou sorry, #1431 has a [mistake](https://github.com/Calinou/scoop-games/issues/1430#issuecomment-2937159498)

Since `melonds.toml` doesn't exist initially, it's created by scoop as a folder, including when you `scoop update --force melonds` with the file _already existing_, apparently. Looks like `melonds.ini` already had special handling for not existing during install.

I modified the manifest locally this time and did `scoop update --force melonds` again. It correctly deleted the invalid _folder_ symlink created by running #1431, and set up the toml file correctly. Melonds functioned normally on its first run after this upgrade, including storing changed settings in the toml file.

image

One thing I'm not sure how to test correctly now is the case of someone with a <1.0 version with content in the .ini file upgrading to 1.0 and this updated manifest preemptively creating the toml file. If melonds' logic to trigger its ini content migration is the absence of the toml file, creating the toml file ahead of time breaks this migration. We could test this and check melonds' code, but it feels wrong to entangle the manifest so hard with remote implementation details.

I would suggest we merge this for now anyway because #1431 left alone is positively breaking.

People who've run #1431 in the past few hours will need to `--force` run this change (or wait for a melonds update), or manually delete the folder symlink. Maybe you know a way to further enhance this but I feel we'd be working a problem that hasn't actually occurred in the wild yet.

Sorry for the mess, I've never worked on scoop packages before.

### DIFF
--- a/bucket/melonds.json
+++ b/bucket/melonds.json
@@ -10,9 +10,14 @@
     "hash": "eda68ed9295bfce4504d9356d3614fbd77648dc3ab1e4ad415f78778f745b33e",
     "installer": {
         "script": [
-            "$FILE = 'melonDS.ini'",
-            "if (!(Test-Path \"$persist_dir\\$FILE\")) {",
-            "    New-Item \"$dir\\$FILE\" -Type File | Out-Null",
+            "if (Test-Path \"$persist_dir\\melonDS.toml\" -PathType Container) {",
+            "    Remove-Item \"$persist_dir\\melonDS.toml\" -Recurse -Force",
+            "}",
+            "$FILES = @('melonDS.ini', 'melonDS.toml')",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

leaumar:
Thanks for the speedy followup 🙂

I think there's still an edgecase where someone will lose their settings during a normal update: the script just creates a blank toml file in `persist/` and symlinks that without copying the content of an _existing_, not-yet-symlinked toml file from the `current/` folder, but I can't work on it all anymore today.

If someone does run into this problem, their toml file is easily restored from the .old folder and they can patch this script too. At least for now a completely fresh install works without problems:

image

PR #1592, opened 2025-12-27
xenia-canary: Modify pre_install script

Mahir007:
Changes made in this PR:
- Ensures 'recent.toml' is created when "AppData" is moved

	This is done because the external folder (created by xenia master) will not contain a 'recent.toml' file but one is needed so that it is persisted correctly.


Calinou (approved):
Thanks!

PR #1629, opened 2026-01-14
Rollback amulet-map-editor manifest to version 0.10.44 and add notes

### LINKED ISSUE #1626: [Bug]: Amulet Map Editor Manifest broken by Licensing Changes
As of version 0.10.45, Amulet [requires a paid license to use](https://www.amuletmc.com/), and the last version that can be downloaded free of charge is 0.10.44, making the Amulet manifest obsolete. I would suggest deprecating/removing the manifest or rolling it back to version 0.10.44. The manifest also does not work due to it being set on a version with no public download links (0.10.47) on GitHub.

bc-185:
Amulet has moved to a paid subscription model for distribution for versions 0.10.45+. This has resulted in autoupdate bots putting invalid links (using the same structure as the links for previous releases) for non-public versions distributed under this new model. I have rolled back the manifest to the last publicly available version (0.10.44) and added a note.

Closes #1626

### DIFF
--- a/bucket/amulet-map-editor.json
+++ b/bucket/amulet-map-editor.json
@@ -1,6 +1,6 @@
 {
     "version": "0.10.44",
-    "description": "Minecraft world editor and converter.",
+    "description": "Minecraft world editor and converter",
     "homepage": "https://www.amuletmc.com",
     "license": "Proprietary",
     "notes": [


Calinou (approved):
Thanks!

PR #1635, opened 2026-01-22
vortex: Add version 1.15.2

### LINKED ISSUE #771: Vortex plugins fail to load
Probably related to #245 

When installing Vortex, all or almost all plugins should work out of the box. But when installing Vortex with Scoop, lots of plugins crash with Failed status

One of the plugins that I was interested in, the one for Skyrim Special Edition - failed with the following error
`Failed to load: Cannot find module 'exe-version'`

I think there's some modules/plugins that should get installed along with the mod manager, but they don't

zeldrisho:
Closes #771, ScoopInstaller/Extras/issues/12050

### DIFF
--- a/bucket/vortex.json
+++ b/bucket/vortex.json
@@ -1,0 +1,34 @@
+{
+    "version": "1.15.2",
+    "description": "Open-source mod manager from Nexus Mods",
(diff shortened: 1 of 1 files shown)


Calinou (approved):
Thanks!

PR #1724, opened 2026-05-11
(chore): set explicit ci workflow permissions

chawyehsu:
Changes synced from other official buckets.

refs:

- https://github.com/ScoopInstaller/Main/commit/d961e02a8284ef6fb1643790567992f28b679c9e
- https://github.com/ScoopInstaller/Extras/commit/8e8a30262fccbfc18801165e32dfb4b3cc761485
- https://github.com/ScoopInstaller/Extras/commit/1052a79abd07743f5b4d1cd9a70c569f0cd4fe92

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on:
   push:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Test
--- a/.github/workflows/excavator.yml
+++ b/.github/workflows/excavator.yml
@@ -1,17 +1,25 @@
+name: Excavator
+
 on:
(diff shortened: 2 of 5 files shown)


chawyehsu:
@Calinou What are your thoughts on having this repository maintained under the ScoopInstaller org since it's listed as one of the official buckets, so that collaborators brought in from ScoopInstaller org can reduce your burden while ensuring it receives timely org-level revisions?

You don't need to implement any actions like repository transfers in addition to this PR until we determine it's feasible.

Calinou:
I agree it would be better to move this bucket to the ScoopInstaller organization, for consistency with other official buckets.

Calinou (approved):

PR #1740, opened 2026-05-27
cs-demo-manager: Fix `pre_install`

woyxiang:

The `Remove-Item *.dll` command in `pre_install` was missing the `$dir\` prefix, causing it to target the current working directory instead of the application's install directory. This could lead to unintended file deletions and the command silently failing.


Calinou (approved):
Thanks!

PR #1745, opened 2026-06-02
ryujinx(-canary): Drop suggest, modify autoupdate & shortcut/bin

Mahir007:
This PR does the following:
 - `ryujinx`:
   - Drop `suggest` property as the suggested item isn't relevant to Ryujinx specifically. It's a more generic tool usable with any Switch emulator and isn't a needed dependency. Additionally, the repo is no longer maintained & has been archived.
   - Updated `autoupdate` url to point at the 'latest' endpoint for the reason of more "correctness".
 - `ryujinx-canary`
   - Same two changes as above.
   - Added available arm64 build.
   - Modified description to specify this is a separate canary version (as can be seen with other emulators in the bucket).
   - Updated shortcuts/bin to again specify the version so one can install both stable and canary releases at the same time.

### DIFF
--- a/bucket/ryujinx-canary.json
+++ b/bucket/ryujinx-canary.json
@@ -14,6 +14,10 @@
         "64bit": {
             "url": "https://git.ryujinx.app/ryubing/canary/releases/download/1.3.308/ryujinx-canary-1.3.308-win_x64.zip",
             "hash": "aedbc0cff468c69d9d27251c5dc8d146111700dcaa1196b96cc91c83cb70f83c"
+        },
+        "arm64": {
+            "url": "https://git.ryujinx.app/ryubing/canary/releases/download/1.3.308/ryujinx-canary-1.3.308-win_arm64.zip",
+            "hash": "90e456b59f12f40da6f3a8c3456699fea8778d2e1cc23f53df2530527ab801ac"
         }
     },
     "extract_dir": "publish",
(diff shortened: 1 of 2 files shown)


Calinou (approved):
Thanks!

bucket/ryujinx-canary.json
```diff
@@ -30,22 +31,28 @@
         "   }",
         "}"
     ],
-    "bin": "Ryujinx.exe",
+    "bin": [
+        "Ryujinx.exe",
+        "ryujinx-canary"
+    ],
```
z-Fng, line 37: Just hit this error. It seems like this change might have introduced a small side effect.
```console
Linking D:\Scoop\apps\ryujinx-canary\current => D:\Scoop\apps\ryujinx-canary\1.3.308
Creating shim for 'Ryujinx'.
Creating shim for 'ryujinx-canary'.
Get-Command: D:\Scoop\apps\scoop\current\lib\install.ps1:191
Line |
 191 |              $bin = (Get-Command $target).Source
     |                      ~~~~~~~~~~~~~~~~~~~
     | The term 'ryujinx-canary' is not recognized as a name of a cmdlet, function, script file, or executable program.
     | Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
Can't shim 'ryujinx-canary': File doesn't exist.
```

Mahir007: oops 😅
i forgot the requirement of needing to use nested square brackets for renaming executables in the bin section. PR #1751 fixes this 🙏🏽

PR #1767, opened 2026-07-08
millennium-installer: Add version 1.12.1

emotion3459:
# Summary
Adds a manifest for the install manager of Millennium, a popular mod framework for the Steam client (basically, Vencord but for steam).
The app is sufficiently popular. 4,000 stars on their [github](https://github.com/SteamClientHomebrew/Millennium) page.

I have a inquiry about what should be done with this app, however.
Millennium stores all of its config files in %STEAM_DIR%/millennium, and ideally we want this to be persisted.
I believe the existing steam manifest could be modified to persist these and it wouldn't cause issues for non-millennium users, but is that okay with you?

 
### DIFF
--- a/bucket/millennium-installer.json
+++ b/bucket/millennium-installer.json
@@ -1,0 +1,31 @@
+{
+    "version": "1.12.1",
+    "description": "A simple standalone program which automates the installation, removal and maintenance of Millennium.",
+    "homepage": "https://steambrew.app",
+    "license": "MIT",
+    "suggest": {
+        "steam": "games/steam"
+    },
+    "architecture": {
+        "64bit": {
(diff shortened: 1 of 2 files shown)


Calinou:
> I believe the existing steam manifest could be modified to persist these and it wouldn't cause issues for non-millennium users, but is that okay with you?

Sounds good to me. Can you make the change to the Steam manifest in this PR?

bucket/millennium-installer.json
```diff
@@ -0,0 +1,31 @@
+{
+    "version": "1.12.1",
+    "description": "A simple standalone program which automates the installation, removal and maintenance of Millennium.",
```
Calinou:
```suggestion
    "description": "Simple standalone program which automates the installation, removal and maintenance of Millennium",
```

Calinou (approved):
Thanks!

PR #1770, opened 2026-07-11
dlss-updater: Update to version 4.3.1 (switch from deprecated zip to MSI)

Recol:
The manifest has been stuck on 3.9.7 because that was the last release to ship a portable zip. The zip artifact was permanently discontinued upstream (Windows Defender false-positives on the PyInstaller bootloader in zip form — see Recol/DLSS-Updater#238), so releases since then ship an MSI and a Flatpak.

The MSI is a plain file-layout package (PyInstaller onedir wrapped by WiX/Briefcase, no registry dependencies, per-user data in %LOCALAPPDATA%), so Scoop's archive-style MSI extraction works: extracting yields `PFiles64\DLSS Updater\` containing `DLSS_Updater.exe` + `_internal\`, verified with an administrative extract of the 4.3.1 MSI.

Changes:
- `url`/`hash`: point at the 4.3.1 MSI release asset
- `extract_dir`: `PFiles64\DLSS Updater`
- `autoupdate` template: `.msi` instead of `.zip`

I'm the upstream author (Recol/DLSS-Updater). The MSI naming scheme (`DLSS.Updater.$version.msi`) is stable across releases, so excavator should keep this current going forward.

Fixes the update path requested in Recol/DLSS-Updater#238.

### DIFF
--- a/bucket/dlss-updater.json
+++ b/bucket/dlss-updater.json
@@ -1,12 +1,12 @@
 {
-    "version": "4.2.3",
+    "version": "4.3.1",
     "description": "A tool that allows you to conveniently update all the DLSS/XeSS/FSR DLLs for the games detected on your system",
     "homepage": "https://github.com/Recol/DLSS-Updater",
     "license": "AGPL-3.0-only",
     "architecture": {
         "64bit": {
-            "url": "https://github.com/Recol/DLSS-Updater/releases/download/V4.2.3/DLSS.Updater.4.2.3.msi",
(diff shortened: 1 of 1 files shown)


Recol:
The Checkver/Autoupdate failures are expected and temporary: the current latest upstream release (V4.3.0) was Linux-only (Flatpak, no MSI asset), so autoupdate's probe for `DLSS.Updater.4.3.0.msi` 404s. V4.3.1 ships shortly for both platforms with the MSI attached.

Calinou (approved):
Thanks!

PR #1777, opened 2026-07-22
wowup-cf: Add version 2.23.0

erayaydin:
Adds a manifest for [WowUp-CF](https://github.com/WowUp/WowUp.CF), the Curseforge enabled build of WowUp.

**Why two Wow-Up versions?**: After WowUp and CurseForge parted ways, the main WowUp app (`wowup`, already [in this bucket](https://github.com/Calinou/scoop-games/blob/master/bucket/wowup.json)) can no longer fetch addons from CurseForge. WowUp.CF is a separate distribution that restores CurseForge support. They're distinct apps with separate repos and installers, so IMO both can coexist in the bucket.

### DIFF
--- a/bucket/wowup-cf.json
+++ b/bucket/wowup-cf.json
@@ -1,6 +1,6 @@
 {
     "version": "2.23.0",
-    "description": "World of Warcraft addon updater with CurseForge support.",
+    "description": "World of Warcraft addon updater with CurseForge support",
     "homepage": "https://wowup.io/",
     "license": {
         "identifier": "GPL-3.0-or-later",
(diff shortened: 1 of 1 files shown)


erayaydin:
> The code here is no longer updated, both clients are now built off of the main repo.

Sorry for the noise, turns out it's not being updated anymore.

erayaydin:
Sorry, my mistake. This still looks valid. The codebase just lives in a different repo, but they still use the releases from there, so I'm reopening. You can also see the `WowUp with CurseForge` button on the [homepage](https://wowup.io/). And they also released latest version `2.23.0`.

bucket/wowup-cf.json
```diff
@@ -0,0 +1,34 @@
+{
+    "version": "2.23.0",
+    "description": "World of Warcraft addon updater with CurseForge support.",
```
Calinou:
```suggestion
    "description": "World of Warcraft addon updater with CurseForge support",
```

Calinou (approved):
Thanks!

PR #1778, opened 2026-07-25
blockbench@5.1.6: fix hash, fix autoupdate, fix architectures

SiriosDev:
- Hashes are now retrieved directly from GitHub (no more sha512 inside release yaml)
- The 32-bit version has been removed because it is no longer distributed (since 4.8, Jul 2023)
- The URLs now point to architecture-specific executables


Calinou (approved):
Thanks!

PR #1780, opened 2026-07-31
pcsx2: Persist textures directory

fadillzzz:
Persist the `textures` directory so they remain usable across different versions. The `pcsx2-dev` package already has this, but for some reason it's missing from the stable version.


Calinou (approved):
Thanks!

